PR #2, opened 2019-10-24
Feature/interface

nairod2000:
The interface between vision and flight.


vision/classes/classes.py
```diff
@@ -0,0 +1,33 @@
...
+    rects.append(rect4)
+    rects.append(rect5)
+
+# information needs to be accessible/ function where generator gets called
+
+
+class Environment:
+
+    def __init__(self, rect):
+        self.rect = rect
```
coledie: Looks like this should not be here.

vision/interface.py
```diff
@@ -0,0 +1,51 @@
...
+    rect1 = Rectangle(4, ",")
+    rect2 = Rectangle(8, ",")
+    rect3 = Rectangle(2, ",")
+    rect4 = Rectangle(6, ",")
+    rect5 = Rectangle(7, ",")
+    rects.append(rect1)
+    rects.append(rect2)
+    rects.append(rect3)
+    rects.append(rect4)
+    rects.append(rect5)
```
coledie:
```suggestion
    rects = [
        Rectangle(4, ","),
        Rectangle(8, ","),
        Rectangle(2, ","),
        Rectangle(6, ","),
        Rectangle(7, ","),
        ]
```

vision/interface.py
```diff
@@ -0,0 +1,51 @@
...
+# information needs to be accessible/ function where generator gets called
+
+
+class Environment:
+
+    def __init__(self):
+        self.rects = []
+
+    def __iter__(self):
+        self.length = len(self.rects)
```
coledie: unused

vision/interface.py
```diff
@@ -0,0 +1,51 @@
...
+                i += 1
+            except ZeroDivisionError:
+                yield None
+
+    def update(self, rects):
+        self.rects = rects
+
+env = Environment()
+for i in env:
+    print(i)
```
coledie: Wrap in if __name__ == '__main__': for import protection.

vision/interface.py
```diff
@@ -0,0 +1,51 @@
...
+
+# information needs to be accessible/ function where generator gets called
+
+
+class Environment:
+
+    def __init__(self):
+        self.rects = []
+
+    def __iter__(self):
```
coledie: Lets add some documentation here.
"""
Function description.

Returns
--------
Rectangle at ..... if values in self.rects else None.
"""

vision/interface.py
```diff
@@ -0,0 +1,51 @@
...
+    rects.append(rect1)
+    rects.append(rect2)
+    rects.append(rect3)
+    rects.append(rect4)
+    rects.append(rect5)
+
+# information needs to be accessible/ function where generator gets called
+
+
+class Environment:
```
coledie: Add a class docstring
"""
Use case
"""

vision/interface.py
```diff
@@ -0,0 +1,51 @@
+#rectangle object
```
coledie: You also need an official unit tests in vision/unit_tests that follows the description given in vision/README.md.

You should give the Environment object a bunch of data normal and some that you think could break it - ensure it works as it should.

coledie: Make sure you include a test where the length of env.rects changes during iteration.

-- also storing self.length is unnecessary because this info is already stored by the interpreter and is just referenced when calling len(). Replace i %= self.length with i %= len(self.rects) and delete other occurrences of self.length.
https://stackoverflow.com/questions/2399835/in-python-what-does-lenlist-do#2399932

vision/interface.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+if __name__ == '__main__':
+    rects = [
+        Rectangle(4, ","),
+        Rectangle(8, ","),
+        Rectangle(2, ","),
+        Rectangle(6, ","),
+        Rectangle(7, ",")
+    ]
```
coledie: Lets combine this with the other if __name__ == '__main__' at the bottom.

vision/interface.py
```diff
@@ -0,0 +1,48 @@
+
+from util import Rectangle
+
+
+"""The Environment class will manage the other objects in the environment"""
```
coledie: This docstring belongs right under the definition of Env
```
class Environment:
    """The Environment class will manage the other objects in the environment"""
```

vision/interface.py
```diff
@@ -0,0 +1,48 @@
+
```
coledie: Docstring here
```
"""
Like this
"""
```

vision/interface.py
```diff
@@ -0,0 +1,48 @@
...
+    ]
+
+
+class Environment:
+
+    #list of rectangles in environment
+    def __init__(self):
+        self.rects = []
+
+    #iterator of the rects list
```
coledie: Documentation belongs under the definition of the function.
```
def __iter__(self):
    """
    Iterator .....
    
    Yields
    ------
    ....
    """
```

vision/interface.py
```diff
@@ -0,0 +1,48 @@
...
+        while True:
+            try:
+                i %= len(self.rects)
+                yield self.rects[i]
+                i += 1
+            except ZeroDivisionError:
+                yield None
+
+    #updates rects list
+    def update(self, rects):
```
coledie: Needs same documentation as __iter__.

vision/rectangle.py
```diff
@@ -0,0 +1,7 @@
+class Rectangle:
```
coledie: Module docstring at top. Class docstring right under definition -- describe parameters of init there too.

PR #4, opened 2019-10-30
Feature/blob

pieperm:


vision/blob/blobfind.py
```diff
@@ -0,0 +1,75 @@
...
+        # create Rectangle and add to list of bounding boxes
+        bbox = Rectangle(top_left_near, top_right_near, bottom_right_near, bottom_left_near, top_left_far, top_right_far, bottom_right_far, bottom_left_far)
+        bounding_boxes.append(bbox)
+
+
+    if logging:
+        im_with_keypoints = cv2.drawKeypoints(image, keypoints, outImage=np.array([]), color=(0, 225, 255), flags=cv2.DRAW_MATCHES_FLAGS_DRAW_RICH_KEYPOINTS)
+        print('Bounding boxes:', bounding_boxes)
+        cv2.imshow("Blobs", im_with_keypoints)
+        cv2.waitKey(0)
```
coledie: Remove the logging variable and everything in this if statement. Lets put this somewhere else though as a nifty utility, maybe a vision/tools or vision/util directory -- generalize it so other people can make use of it.

vision/blob/blobfind.py
```diff
@@ -0,0 +1,75 @@
...
+def configure_params():
+    params = cv2.SimpleBlobDetector_Params()
+
+    params.filterByArea = True
+    params.maxArea = constants.MAX_AREA
+
+    params.minThreshold = constants.MIN_THRESHOLD
+    params.maxThreshold = constants.MAX_THRESHOLD
+
+    return params
```
coledie: Lets make the config a .json so we dont need a custom parser and itll be more accessible to others trying to use the config or change it. See my comments on the constants.py.

vision/blob/blobfind.py
```diff
@@ -0,0 +1,75 @@
...
+
+        # With depth, these will be calculated differently
+        top_left_far = top_left_near
+        top_right_far = top_right_near
+        bottom_right_far = bottom_right_near
+        bottom_left_far = bottom_left_near
+
+        # create Rectangle and add to list of bounding boxes
+        bbox = Rectangle(top_left_near, top_right_near, bottom_right_near, bottom_left_near, top_left_far, top_right_far, bottom_right_far, bottom_left_far)
+        bounding_boxes.append(bbox)
```
coledie: Will need to update w/ new rect class from interface.

vision/blob/blobfind.py
```diff
@@ -0,0 +1,75 @@
...
+    for keypoint in keypoints:
+        # find center and radius of keypoint
+        x_coord, y_coord = keypoint.pt[0], keypoint.pt[1]
+        radius = keypoint.size / 2
+
+        # calculate coordinates for bounding box of each blob
+        top_left_near = (x_coord - radius, y_coord - radius, 0)
+        top_right_near = (x_coord + radius, y_coord - radius, 0)
+        bottom_right_near = (x_coord + radius, y_coord + radius, 0)
+        bottom_left_near = (x_coord - radius, y_coord + radius, 0)
```
coledie: calculate x - r, x + r, ... once and pass into each for readability.

vision/blob/blobfind.py
```diff
@@ -0,0 +1,75 @@
...
+
+    params = configure_params()
+    blob_detector = cv2.SimpleBlobDetector_create(params)
+
+    keypoints = blob_detector.detect(image)
+
+    bounding_boxes = []
+    for keypoint in keypoints:
+        # find center and radius of keypoint
+        x_coord, y_coord = keypoint.pt[0], keypoint.pt[1]
```
coledie: Maybe center_x and center_y for readability.

vision/blob/blobfind.py
```diff
@@ -0,0 +1,75 @@
+import cv2
```
coledie: \"""
Docstring at top
"""

vision/Blobdetection/simpleDetection.py
```diff
@@ -0,0 +1,70 @@
+
```
coledie: Lets put this code in vision/camera! I can have someone format all of the types of camera streams so they are generic interfaces for all the vision algorithms. Rename it to webcam.py or something.

vision/blob/blob.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+    def __init__(self, top_left_near, top_right_near, bottom_right_near, bottom_left_near, top_left_far, top_right_far, bottom_right_far, bottom_left_far):
+        self.top_left_near = top_left_near
+        self.top_right_near = top_right_near
+        self.bottom_right_near = bottom_right_near
+        self.bottom_left_near = bottom_left_near
+        self.top_left_far = top_left_far
+        self.top_right_far = top_right_far
+        self.bottom_right_far = bottom_right_far
+        self.bottom_left_far = bottom_left_far
```
coledie: If you pull feature/interface, you can use their rectangle class.

vision/blob/blob.py
```diff
@@ -0,0 +1,53 @@
...
+
+class Blob:
```
coledie: I think you should add a object_type parameter to rectangle rather than creating a wrapper.

vision/blob/blobfind.py
```diff
@@ -0,0 +1,105 @@
...
+"""
+import os
+import sys
+sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__)))))
+import cv2
+from vision.blob.blob import Rectangle
+import json
+
+def import_params():
```
coledie: We need a unit test for this.

vision/blob/blobfind.py
```diff
@@ -0,0 +1,105 @@
...
+        for category in config:
+            category_enabled = config[category]['enable']
+            if category_enabled:
+                if hasattr(params, category):
+                    setattr(params, category, config[category]['enable'])
+                for attr in config[category]:
+                    if hasattr(params, attr):
+                        setattr(params, attr, config[category][attr])
+
+        print(params.minThreshold)
```
coledie: Remove this.

vision/blob/blobfind.py
```diff
@@ -0,0 +1,105 @@
...
+    params: SimpleBlobDetector_Params
+        blob detector params object
+    """
+    def __init__(self, image, params=None):
+        self.image = image
+        self.keypoints = []
+        if params is None:
+            self.params = import_params()
+        else:
+            self.params = params
```
coledie:
```
self.params = import_params() if params is None else params
```

vision/blob/blobfind.py
```diff
@@ -0,0 +1,105 @@
...
+        self.params = params
+
+    """
+    Detects blobs in the image provided in the constructor
+
+    Returns
+    -------
+    list[Rectangle]
+        a list of bounding boxes represented as Rectangles, each with 8 (x, y, z) coordinates
+    """
```
coledie: Put documentation under function definition.

vision/blob/blobfind.py
```diff
@@ -0,0 +1,105 @@
...
+        if params is None:
+            self.params = import_params()
+        else:
+            self.params = params
+
+    def update_image(self, image):
+        self.image = image
+
+    def update_params(self, params):
+        self.params = params
```
coledie: Lets use properties for these ie pythons take on getters and setters.
```
@property
def image(self):
    """
    This function is called every time self.image is run.
    """
    return self._image  # You shouldn't need to rename any instances of image to _image, the getter and setters will work.

@image.setter
def image(self, value):
    """
    Defines behavior of self.image = value.
    """
    if not instance(value, np.ndarray):
        raise ValueError("Need an image")
    self._image = image
```
Same for params

vision/blob/blobfind.py
```diff
@@ -0,0 +1,105 @@
...
+    """
+    Detects blobs in the image provided in the constructor
+
+    Returns
+    -------
+    list[Rectangle]
+        a list of bounding boxes represented as Rectangles, each with 8 (x, y, z) coordinates
+    """
+    def find(self):
+        blob_detector = cv2.SimpleBlobDetector_create(self.params)
```
coledie: This belongs in the __init__. If the params are updates, the params setter should create a new instance of this.

vision/unit_tests/test_blobbing.py
```diff
@@ -0,0 +1,46 @@
...
+        expected_blobs = {
+            "apple.jpg": 1,
+            "legos.jpg": 30,
+            "MyBeach.png": 1,
+            "oranges.png": 1,
+            "sampleobj.png": 1
+        }
+        vision_folder = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+
+        for i, img in enumerate(os.listdir('vision/blob/samples')):
```
coledie: We can refactor this to 
```
for filename, expected in expected_blobs.items():
```
I think that will clean up your code below and prevent it from breaking with new files in the samples directory.

vision/util/blob_logger.py
```diff
@@ -0,0 +1,36 @@
...
+    from blob.blobfind import BlobFinder
+
+def log_blobs(blob_finder):
+    """
+    Shows the detected blobs in an image and print the bounding boxes
+
+    Parameters
+    ----------
+    blob_finder: BlobFinder
+        BlobFinder object that contains an image and potentially params
```
coledie: Lets change this to a list of rectangles, which the BlobFinder outputs. This will make it generic to anyone's algorithm. In the if __name__ == '__main__' of blob you could run blob finder and have it print this.

May need to pass in an image too.

vision/util/blob_logger.py
```diff
@@ -0,0 +1,36 @@
+"""Utility for logging blobs detected in an image"""
+
+import cv2
+import numpy as np
+try:
+    from vision.blob.blobfind import BlobFinder
+except ImportError:
+    from blob.blobfind import BlobFinder
+
+def log_blobs(blob_finder):
```
coledie: Rename to show_blobs so that people will know what they are in for.

pieperm:
**Note**: We're going to also have to move the images in `blob/samples` to `vision/vision_images/blob` and fix file paths accordingly

PR #5, opened 2019-11-08
Feature/modulecheckers

fallscameron01:
Added ModuleInFrame.py. This file contains the function ModuleInFrame that will determine if the module is in an image. It takes an image path as a parameter and returns a bool, which is True if the module is in the image and false otherwise.


vision/module_in_frame/ModuleInFrame.py
```diff
@@ -0,0 +1,68 @@
+"""
+This file contains the ModuleInFrame function to detect if the module is in an image
+"""
+
+import cv2
+import numpy as np
+
+def ModuleInFrame(imgPath):
```
coledie: Refactor so that the image comes in as a numpy array, in production we will not be saving images.

vision/module_in_frame/ModuleInFrame.py
```diff
@@ -0,0 +1,68 @@
...
+    """
+
+    # Ignore numpy warnings
+    np.seterr(all="ignore")
+
+    # Get target image
+    img = cv2.imread(imgPath)
+
+    # Remove depth channel
+    img.reshape((-1, 3))
```
coledie: This will string out and jumble up the image instead of removing the depth. -- also this is running the function but not overriding the data in img, which is why the code still works.

```
img = img[:, :, :3]  # This forces the image to 3 channels.
```

vision/module_in_frame/ModuleInFrame.py
```diff
@@ -0,0 +1,68 @@
...
+
+    # Grayscale
+    gray = cv2.cvtColor(src=img, code=cv2.COLOR_RGB2GRAY)
+
+    # Guassian Blur
+    blur = cv2.GaussianBlur(src=gray, ksize=(5,5), sigmaX=0)
+
+    # Laplacian Transform
+    laplacian = cv2.Laplacian(src=blur, ddepth=cv2.CV_8U, ksize=3)
+    laplacian = np.uint8(np.around(laplacian))
```
coledie: Just converting to int will have the same effect as np.around.

```
laplacian = np.uint8(laplacian)
```

vision/module_in_frame/ModuleInFrame.py
```diff
@@ -0,0 +1,68 @@
...
+    img.reshape((-1, 3))
+
+    # Grayscale
+    gray = cv2.cvtColor(src=img, code=cv2.COLOR_RGB2GRAY)
+
+    # Guassian Blur
+    blur = cv2.GaussianBlur(src=gray, ksize=(5,5), sigmaX=0)
+
+    # Laplacian Transform
+    laplacian = cv2.Laplacian(src=blur, ddepth=cv2.CV_8U, ksize=3)
```
coledie: We should see how these k sizes respond to large/high resolution images. They may need to be dynamic to the size of the image.

vision/module_in_frame/ModuleInFrame.py
```diff
@@ -0,0 +1,68 @@
...
+    laplacian = np.uint8(np.around(laplacian))
+
+    # Hough Circle Detection
+    circles = cv2.HoughCircles(image=laplacian, method=cv2.HOUGH_GRADIENT, dp=1, minDist=8, param1=50, param2=40, minRadius=0, maxRadius=50)
+    circles = np.uint16(np.around(circles))
```
coledie: Just converting to int will have the same effect as np.around.

```
circles = np.uint16(circles)
```

vision/module_in_frame/ModuleInFrame.py
```diff
@@ -0,0 +1,68 @@
...
+
+    # Guassian Blur
+    blur = cv2.GaussianBlur(src=gray, ksize=(5,5), sigmaX=0)
+
+    # Laplacian Transform
+    laplacian = cv2.Laplacian(src=blur, ddepth=cv2.CV_8U, ksize=3)
+    laplacian = np.uint8(np.around(laplacian))
+
+    # Hough Circle Detection
+    circles = cv2.HoughCircles(image=laplacian, method=cv2.HOUGH_GRADIENT, dp=1, minDist=8, param1=50, param2=40, minRadius=0, maxRadius=50)
```
coledie: We will need to parameterize this soon, though someone is already writing some good code to do so and we can just copy paste it.

vision/module_in_frame/ModuleInFrame.py
```diff
@@ -0,0 +1,68 @@
...
+        if(val > 1):
+            parallels += val
+        for (key2, val2) in slopes.items():
+            if(abs(key - key2) < .01 and key != key2):
+                parallels += 1
+
+    # Determine if module is in frame
+    isInFrame = False
+    if(parallels > 300):
+        isInFrame = True
```
coledie:
```
isInFrame = parallels > 300
```

vision/unit_tests/module_in_frame/UnitTest.py
```diff
@@ -0,0 +1,26 @@
...
+        Returns
+        -------
+        ndarray[bool] which images the module was detected in
+        """
+
+        results = []
+        expected_results = [True, True, True, True]
+        for pic in ["Block1.png", "Block2.png", "Block3.jpg", "Block4.jpg"]:
+            results.append(mif(pic))
+        self.assertListEqual(results, expected_results)
```
coledie: Lets make expected results into a dictionary, {'filename': value}.
```
for pic in expected_results.keys():
    results.append(mif(pic))
self.assertListEqual(results, list(expected_results.keys()))
```

vision/unit_tests/module_in_frame/UnitTest.py
```diff
@@ -0,0 +1,26 @@
...
+        """
+        Testing ModuleInFrame
+
+        Returns
+        -------
+        ndarray[bool] which images the module was detected in
+        """
+
+        results = []
+        expected_results = [True, True, True, True]
```
coledie: Lets add some extra non-module images to see if the algorithm gives false positives.

vision/module_in_frame/ModuleInFrame.py
```diff
@@ -0,0 +1,68 @@
...
+                        slopes[m] = 0
+                    slopes[m] += 1
+
+    # Counting the number of parallel slopes
+    parallels = 0
+    for(key, val) in slopes.items():
+        if(val > 1):
+            parallels += val
+        for (key2, val2) in slopes.items():
+            if(abs(key - key2) < .01 and key != key2):
```
coledie:
```
key != key2 and abs(key - key2) < .01
``` 
The abs(..) will not execute if the keys are equal, helping in with speed.

vision/module_in_frame/ModuleInFrame.py
```diff
@@ -0,0 +1,68 @@
...
+            for iCircle in circles:
+                for (iX, iY, iR) in iCircle:
+                    m = (iY - y) / (iX - x)
+                    if(m not in slopes):
+                        slopes[m] = 0
+                    slopes[m] += 1
+
+    # Counting the number of parallel slopes
+    parallels = 0
+    for(key, val) in slopes.items():
```
coledie: Rename key, val to slope, count or something more readable.

vision/module_in_frame/ModuleInFrame.py
```diff
@@ -0,0 +1,68 @@
...
+        if(val > 1):
+            parallels += val
+        for (key2, val2) in slopes.items():
+            if(abs(key - key2) < .01 and key != key2):
+                parallels += 1
```
coledie: You should define what parallel means -- if 2 lines are parallel should you add 1 or 2 to parallels to the counter, if 3 lines are ....
It appears that if 2 lines have exactly the same slope, you increment parallels by 2 once. If they are close within a threshold you add one to the counter twice, which is legit.
BUT
If 3 lines are parallel within a threshold you add 1 to the counter 6 times in the abs < .1.
If 2 lines have exactly the same slope and they are parallel within a threshold to another, you add 2 to the counter in the val > 1 and then add 1 to the counter twice in the abs < .1.
If 3 lines have exactly the same slope you add 3 to the counter once in the val > 1.

I would think about this more AFTER you settle on a data structure for slopes(see comment above).

vision/module_in_frame/ModuleInFrame.py
coledie: Make the .01 a SLOPE_THRESHOLD(or another cool name) constant/defaulted parameter.

vision/module_in_frame/ModuleInFrame.py
```diff
@@ -0,0 +1,68 @@
...
+                    if(m not in slopes):
+                        slopes[m] = 0
+                    slopes[m] += 1
+
+    # Counting the number of parallel slopes
+    parallels = 0
+    for(key, val) in slopes.items():
+        if(val > 1):
+            parallels += val
+        for (key2, val2) in slopes.items():
```
coledie: These for loops will iterate over every value-value pair twice! If we turn slopes into a list, we can do something similar to the following
```
for i, slope1 in slopes:
    for slope2 in slopes[i+1:]:
```
This cuts the number of for loop iterations in half, but may be a little ugly with a dictionary.

vision/module_in_frame/ModuleInFrame.py
```diff
@@ -0,0 +1,68 @@
...
+    laplacian = np.uint8(np.around(laplacian))
+
+    # Hough Circle Detection
+    circles = cv2.HoughCircles(image=laplacian, method=cv2.HOUGH_GRADIENT, dp=1, minDist=8, param1=50, param2=40, minRadius=0, maxRadius=50)
+    circles = np.uint16(np.around(circles))
+
+    # Finding slopes between the circles
+    slopes = { }
+    for circle in circles:
+        for (x, y, r) in circle:
```
coledie: Is circle a single (x, y, r) tuple or is it a set of them? If it's only one then lets remove the for loops. If its a set of tuples lets not iterate through every single one for speed concerns ie
```
for (x, y, r) in circle[::3]:
```
This looks at every 3rd tuple in circle.

vision/module_in_frame/ModuleInFrame.py
```diff
@@ -0,0 +1,68 @@
...
+            for iCircle in circles:
+                for (iX, iY, iR) in iCircle:
+                    m = (iY - y) / (iX - x)
+                    if(m not in slopes):
+                        slopes[m] = 0
+                    slopes[m] += 1
```
coledie: I think we should either make a greater use of slopes as a dictionary or change data structures to a list. 
As it stands now the code could be simplified & sped up by switching to a list ie [m1, m2 ...]. We may even be able to use numpy to make it stupid fast. 
If you would like to continue using this as a dictionary, I'd recommend grouping data into buckets while adding them to the dict ie round m to the nearest .5 then do slopes[m] += 1. This way you can skip the entire slope comparison/parallel counting step. This could be extremely fast -- as fast or faster than numpy, especially if the code is cythonized! Though I'm just speculating.

vision/module_in_frame/ModuleInFrame.py
```diff
@@ -0,0 +1,68 @@
...
+
+    # Hough Circle Detection
+    circles = cv2.HoughCircles(image=laplacian, method=cv2.HOUGH_GRADIENT, dp=1, minDist=8, param1=50, param2=40, minRadius=0, maxRadius=50)
+    circles = np.uint16(np.around(circles))
+
+    # Finding slopes between the circles
+    slopes = { }
+    for circle in circles:
+        for (x, y, r) in circle:
+            for iCircle in circles:
```
coledie: This does every circle-circle comparison twice.
```
for i, circle1 in circle:
    for circle2 in circle[i+1:]:
```

vision/unit_tests/module_in_frame/UnitTest.py
```diff
@@ -0,0 +1,26 @@
+"""
```
coledie: Let's add a unit tests to ensure that every circle is considered in the slope calculating, ie make an image of circles and ensure it has the expected slope values between circles. You will probably need to sort the lists before comparing them.

vision/module_in_frame/ModuleInFrame.py
```diff
@@ -0,0 +1,67 @@
...
+            if (not np.isnan(m)) and (not np.isinf(m)) and (x != iX and y != iY):
+                slopes = np.append(slopes, m)
+    
+    # Converting slopes to degrees
+    slopes = np.degrees(np.arctan(slopes))
+    
+    # Bucket sorting slopes to group parallels
+    upper_bound = np.amax(slopes)
+    lower_bound = np.amin(slopes)
+    BUCKET_MODIFIER = 1 # Changes how many buckets are in the range
```
coledie: Lets put these constants at the top of the file to make them really easy to modify.

vision/module_in_frame/ModuleInFrame.py
```diff
@@ -0,0 +1,67 @@
...
+    np.seterr(all="ignore")
+
+    # Remove depth channel
+    img = img[:, :, :3]
+
+    # Grayscale
+    gray = cv2.cvtColor(src=img, code=cv2.COLOR_RGB2GRAY)
+
+    # Guassian Blur
+    blur = cv2.GaussianBlur(src=gray, ksize=(5,5), sigmaX=0)
```
coledie: I think ksize belongs as a constant at the top with BUCKET_MODIFIER. Just
```
    blur = cv2.GaussianBlur(src=gray, ksize=(BLUR_SIZE, BLUR_SIZE), sigmaX=0)
```

vision/unit_tests/module_in_frame/test_modulecheck.py
```diff
@@ -0,0 +1,33 @@
+"""
+This file is used for testing ModuleInFrame
+"""
+
+import os, sys
+sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__)))))
```
coledie: Lets move the test_moduleckeck file directly to the unit_tests directory and remove the module_in_frame folder here. You will be able to move this second sys.path.insert then.

PR #6, opened 2019-11-11
Feature/module

FaolanCody:
Implementation and Unittest for the ModuleKMeans class. Allows for color distinction, useful as a baseline for module detection.


vision/Unittest.py
```diff
@@ -0,0 +1,38 @@
...
+        Test images must be put in a file called images
+        """
+        
+        for pic in ["images/blocks1.jpg",
+                    "images/blocks2.jpg",
+                    "images/pylon.jpg"]:
+            img = mkm(pic)
+            img.applyKMeans(3)
+            channels = [0, 1]
+            img.displayBinary(channels)
```
coledie: Both of these tests require a human to interpret the results, which is okay, but should not be done here as a unit test. Id recommend putting both of these functions in the `if __name__ == '__main__':` of the  KMeans file. We do want to run a unit test on the clustering algorithm and make sure it responds to parameter changes as it should, but without a tool to generate a goal output for each image this is hard. For now you should find an image that your algorithm works well on and save the output, the unit test should ensure that future output resembles it within some error margin. This way whoever creates the tool for us will have something to work with.

vision/Unittest.py
```diff
@@ -0,0 +1,38 @@
+import unittest
```
coledie: Move this folder to vision/unit_tests and rename it to test_kmeans.py

vision/module/ModuleKMeans.py
```diff
@@ -0,0 +1,126 @@
...
+    ---------
+    imgPath: string
+        Reads in the image with this name to be used
+
+    channel_weights: float array
+        Decides the weight of each channel in the image
+    """
+
+    def __init__(self, imgPath, channel_weights=[1., 1., 1.]):
+        self.img = cv2.imread(imgPath)
```
coledie: The init should initialize everything that will be reused in the function. The apply KMeans function should take in an image and process it. This way we create one object and can reuse it for every image.

coledie: You should not be taking in an image path, just a straight numpy array.

vision/module/ModuleKMeans.py
```diff
@@ -0,0 +1,126 @@
...
+        self.pixelData = self.img.reshape((-1, 3))
+        self.pixelData = np.float32(self.pixelData)
+        self.pixelData *= channel_weights
```
coledie: Lets move all this to applyKMeans and make channel weights a parameter of that.

coledie: What if we do this to make these lines robust to 3 and 4 channel images.
```
        if len(self.img.shape) == 2 and len(channel_weights) > 1:
            raise ValueError(f"Image incorrect shape, expected: {len(channel_weights)} got 1")

        if len(self.img.shape) == 3 and self.img.shape[-1] == len(channel_weights):
            raise ValueError(f"Image incorrect shape, expected: {len(channel_weights)} got {self.img.shape[-1]}")

        self.pixelData = self.img.reshape((-1, len(channel_weights)))
        self.pixelData = np.float32(self.pixelData)
        self.pixelData *= channel_weights
```

vision/unit_tests/test_kmeans.py
```diff
@@ -0,0 +1,26 @@
...
+        pass
+
+    def test_binary(self):
+        """
+        Testing Binary KMeans Display
+
+        Test images must be put in a file called images
+        """
+
+        pass
```
coledie: Just replace everything under the class definition with pass
ie
```
class TestKMeans(unittest.TestCase):
    pass
```

vision/module/ModuleKMeans.py
```diff
@@ -0,0 +1,126 @@
...
+        self.defaultCriteria = (cv2.TERM_CRITERIA_EPS +
+                                cv2.TERM_CRITERIA_MAX_ITER,
+                                10, 1.0)
```
coledie: You can move this to the applyKMeans function ie
```
        if criteria == 0:
            criteria = (cv2.TERM_CRITERIA_EPS +
                                cv2.TERM_CRITERIA_MAX_ITER,
                                10, 1.0)

File: vision/module/requirements.txt [hunk 7]
```diff
@@ -0,0 +1,2 @@
+numpy
+opencv
```
coledie: opencv-python

PR #12, opened 2020-01-17
Feature/pipeline

nairod2000:


vision/pipeline.py
```diff
@@ -35,8 +35,8 @@ def run_algorithm(self):
         """
         Method that takes the given video file and environment, and updates the environment with detected blobs.
         """
-        for self.vid_len, (depth_image, color_image) in enumerate(ReadBag(self.vid_file)):
-            if self.vid_len == 98:
+        for self.alg_time, (depth_image, color_image) in enumerate(ReadBag(self.vid_file)):
+            if self.alg_time == 98:
```
coledie: I think you meant to change this to
```
for self.vid_len, (depth_image, color_image) in enumerate(ReadBag(self.vid_file)):
            if self.vid_len == self.alg_time:
```

PR #14, opened 2020-01-17
adding multiprocessing

aydenmason:
Multiprocessing functionality added so that the vision and flight code can read/write to the same object.


vision/interface.py
```diff
@@ -1,7 +1,7 @@
 
 """The Environment class will manage the other objects in the environment"""
```
coledie: This looks good but is definitely needs a unit test, ensuring no crashes and no data loss. Have 2-3 agents trying to update at the same time and maybe 1 agent trying to read from.

vision/interface.py
```diff
@@ -1,7 +1,7 @@
...
-
-
+import multiprocessing as mp 
+from functools import partial
 class Environment:
 
     #list of rectangles in environment
```
coledie: Let's switch this comment to a docstring, I'm not sure if code editors will pick up on this.

vision/interface.py
```diff
@@ -23,3 +23,18 @@ def __iter__(self):
     #updates rects list
```
coledie: Lets switch this comment to a docstring and also put it right under the definition
ie
```python
def update(self, bounding_boxes):
    """
    description

    Parameters
    ------------
    bounding_boxes: type
         description
```

vision/interface.py
```diff
@@ -23,3 +23,18 @@ def __iter__(self):
...
     def update(self, bounding_boxes):
         self.bounding_boxes = bounding_boxes
+
+    #This is the functiont that the teams should call
+    #safe_update handles all multiprocessing options
+    #CAUTION: may not run only work on Linux/MacOS
```
coledie: Switch to docstring, move under definition and describe the bounding box parameter for consistency and to make sure code editors pick up on it.

vision/interface.py
```diff
@@ -23,3 +23,18 @@ def __iter__(self):
...
     def update(self, bounding_boxes):
         self.bounding_boxes = bounding_boxes
+
+    #This is the functiont that the teams should call
+    #safe_update handles all multiprocessing options
+    #CAUTION: may not run only work on Linux/MacOS  
+    def safe_update(self, bounding_boxes):
+        pool = mp.Pool()
+        m = mp.Manager()
+        l = m.Lock()
```
coledie: Can we define all or some of these in the __init__ and reuse?

coledie:
Upon further inspection it looks like this would not be helpful to make the interface shareable between threads. This allows the update function to be distributed among processes. The goal is to be able to share the Environment class among separate processes in a memory safe way.

PR #18, opened 2020-01-29
Feature/pylon

skylerwinkelman:
Converts an image to hsv and detects red within a range.  The idea is to detect the perfect shade of red and not any other red objects that could be in the image.


vision/pylon/classify.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+
+    cv2.imshow("Frame", frame)
+    cv2.imshow("Detect",combined)
+
+    key = cv2.waitKey(1)
+    if key == 27:
+        break
+'''
```
coledie: If this code is not used lets remove it.

vision/pylon/classify.py
```diff
@@ -0,0 +1,60 @@
...
+
+#Repeat with white mask
+lowWhite = np.array([0,0,0])
+highWhite = np.array([1,255,255])
+whiteMask = cv2.inRange(hsv, lowWhite, highWhite)
+white = cv2.bitwise_and(img, img, mask=whiteMask)
+
+combined = cv2.bitwise_or(white, red)
+
+edges_high_thresh = cv2.Canny(img, 60, 120)
```
coledie: Lets split this code up into two sections
1. The classifier function, this should take in an already loaded image then return whether thy pylon is in the image or not, this may include commented out debugging tools.
2. The manual debbuging code, this should be present in an  - if __name__ == '__main__': block at the bottom of the file. Thus it will only be run when the user manually calls this file and not every time it is imported. This should include reading in a sample image, then calling the function and displaying the output as you do.

vision/pylon/detect_pylon.py
```diff
@@ -0,0 +1,28 @@
...
+    pylon = False
+    for x in range(0, red_mask.shape[0]):
+        for y in range(0, red_mask.shape[1]):
+            if red_mask[x, y] != 0:
+                pylon = True
+
+    return pylon
+
+image = cv2.imread("sim_pylon2.png")
+print(detect_red(image))
```
coledie: Lets put all this code into an if name equals main block to prevent it from running when it should not, ie
```python
if __name__ == '__main__':
    ...
```

vision/pylon/detect_pylon.py
```diff
@@ -0,0 +1,28 @@
+import cv2
```
coledie: Lets add a docstring to tell a reader what the file does
ie
```python
"""
Description here
"""
```

vision/pylon/detect_pylon.py
```diff
@@ -0,0 +1,28 @@
...
+
+# detect_red detects all pixels that fall in a certain range in an image
+# then outputs the original image and detected mask
+def detect_red(image):
+    # Convert the image from BGR to HSV
+    hsv = cv2.cvtColor(image, cv2.COLOR_BGR2HSV)
+
+    # Set the lower and upper color limits
+    lower_red = np.array([50, 150, 25])
+    upper_red = np.array([255, 255, 120])
```
coledie: We should make these into constants -- ie variable names to all caps, and define them outside of the function definition. This way they can be easily changed by anyone.

vision/pylon/detect_pylon.py
```diff
@@ -0,0 +1,28 @@
...
+    # and replaces them with white.  Everything else is replaced with black
+    red_mask = cv2.inRange(hsv, lower_red, upper_red)
+
+    # Run through each pixel in the mask.  If one is white, then
+    # we detected part of the pylon
+    pylon = False
+    for x in range(0, red_mask.shape[0]):
+        for y in range(0, red_mask.shape[1]):
+            if red_mask[x, y] != 0:
+                pylon = True
```
coledie: We should break out of both of these for loops immediately if it is known the pylon is in here. return True may suffice also.

vision/unit_tests/test_pylon.py
```diff
@@ -0,0 +1,80 @@
...
+import sys
+import json
+import cv2
+
+parent_dir = os.path.dirname(os.path.abspath(__file__))
+gparent_dir = os.path.dirname(parent_dir)
+ggparent_dir = os.path.dirname(gparent_dir)
+sys.path += [parent_dir, gparent_dir, ggparent_dir]
+
+# from vision.pylon.detect_pylon import import_params
```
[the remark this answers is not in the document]
skylerwinkelman: It says "unresolved reference 'vision'" then "unresolved reference 'detect_red.'"
I'm not sure how to access the parent folders like vision and vision.pylon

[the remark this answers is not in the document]
coledie: Are you running code from vision/?

vision/pylon/classify.py
```diff
@@ -0,0 +1,32 @@
...
+    highWhite = np.array([1,255,255])
+    whiteMask = cv2.inRange(hsv, lowWhite, highWhite)
+    white = cv2.bitwise_and(img, img, mask=whiteMask)
+
+    if __name__ == '__main__':
+        combined = cv2.bitwise_or(white, red)
+        edges_high_thresh = cv2.Canny(img, 60, 120)
+        cv2.imshow('Edges',edges_high_thresh)
+        cv2.imshow('combined', combined)
+        cv2.waitKey(0)
```
coledie: Lets add a comment that says this is debugging code.

vision/pylon/classify.py
```diff
@@ -0,0 +1,32 @@
...
+
```
coledie: The function needs to return something, for a classifier it can just be true or false.

Issue #16, opened 2020-01-28
Blob finder take image as parameter of find


pieperm:
Blob finder to take image as parameter of  find function -- not saved in class.

Issue #15, opened 2020-01-28
Move import_params to its own file


pieperm:
Move import_params to a separate file in util/. Make generic to param struct. BlobFinder object take config dictionary as parameter rather than param object.

PR #63, opened 2020-01-30
Feature/blob

pieperm:
Add blob timer; refactor import params as its own utility


.gitignore
```diff
@@ -4,6 +4,7 @@ vision_videos/
 *.jpg
 *.png
 *.bag
+*.ipynb
```
coledie: This should not be ignored, it is runnable code. Though you should reset the output of notebooks before commiting.

vision/util/import_params.py
```diff
@@ -0,0 +1,27 @@
+import cv2
```
coledie: The unit test for this needs to be updated.

vision/blob/blob_timer.py
```diff
@@ -0,0 +1,26 @@
...
+with open(config_filename, 'r') as config_file:
+    config = json.load(config_file)
+
+for img in os.listdir(img_folder):
+    if img[-4:] not in ['.png', '.jpg']:
+        continue
+
+    image = cv2.imread(os.path.join(img_folder, os.fsdecode(img)))
+    blob_finder = BlobFinder(image, params=import_params(config))
+    print(img, image.shape, timeit.timeit(blob_finder.find, number=10) / 10)
```
coledie: Lets put this in an if __name__ == '__main__': block.

coledie:
Resolves #16  Resolves #15

PR #64, opened 2020-01-30
Feature/find image param

pieperm:
**Important:** This pull request contains contents from PR #63 `feature/blob` so that PR should be resolved before this one.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ vision_videos/
 *.jpg
 *.png
 *.bag
+*.ipynb
 
 # Byte-compiled / optimized / DLL files
 __pycache__/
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -4,6 +4,7 @@ vision_videos/
 *.jpg
 *.png
 *.bag
+*.ipynb
```
coledie: This should not be in the gitignore.

vision/util/import_params.py
```diff
@@ -0,0 +1,27 @@
+import cv2
+
```
coledie: Need to update the pull request for this.

PR #80, opened 2020-02-12
Rename vision/blob to vision/obstacle

### LINKED ISSUE #59: Rename blob detection to obstacle detection.
Everything w/ blob in name ie blob/, test_blobbing .... except actual blob detection algorithm

pieperm:


coledie (approved):
No more blobs!

PR #81, opened 2020-02-12
Modified blobfind and test_blobbing to use 2 vertices instead of 3

### LINKED ISSUE #58: Bounding box to 2d rectangle.
Bounding box use 2d rectangle(4 vertices) with a depth parameter. Update unit tests and algorithms accordingly.

skylerwinkelman:


vision/blob/blobfind.py
```diff
@@ -80,18 +80,18 @@ def find(self, image):
...
+            bottom_right_near = (pos_dx, pos_dy)
+            bottom_left_near = (neg_dx, pos_dy)
 
             # With depth, these will be calculated differently
-            top_left_far = top_left_near
+            """top_left_far = top_left_near
             top_right_far = top_right_near
             bottom_right_far = bottom_right_near
-            bottom_left_far = bottom_left_near
+            bottom_left_far = bottom_left_near"""
```
coledie: It appears these are unused so we can delete them.

vision/unit_tests/test_blobbing.py
```diff
@@ -107,12 +107,12 @@ def test_annotation_accuracy(self):
                     ax1, ay1, ax2, ay2 = [int(annotation_bounding_box.find(param).text) for param in ['xmin', 'ymin', 'xmax', 'ymax']]
 
                     for bounding_box in bounding_boxes:
+                        print("VERTICES", bounding_box.vertices)
```
coledie: Lets delete this print statement since it was only used for debugging..

coledie (changes requested):
I like it

coledie (approved):

PR #110, opened 2020-02-26
Add unit test for camera.

### LINKED ISSUE #98: Implement unit tests for all camera types.

coledie:


vision/unit_tests/test_camera.py
```diff
@@ -10,15 +10,103 @@
 import unittest
 from unittest.mock import patch
 
-from vision.camera import *
+import numpy as np
+import airsim
```
ClayMav, line 14: Note: will this cause problems for us being able to run our code being that this library is strictly required and we may be connected to a real camera

ClayMav (approved):

vision/unit_tests/test_camera.py
coledie: There is three separate camera types, two use realsense and one uses airsim.

PR #114, opened 2020-02-27
Detecting the pylon with red threshold works

### LINKED ISSUE #103: Make pylon robust to noise by requiring # red pixels to be greater than threshold.
Currently if any single pixel is red enough, the algorithm assumes the pylon is in image. We can make this more robust by forcing so many pixels to be red for this.

andrewhthomas16:


vision/pylon/detect_pylon.py
```diff
@@ -36,15 +39,19 @@ def detect_red(color_image, depth_image):
...
-    # Run through each pixel in the mask.  If one is white, then
-    # we detected part of the pylon
+    # Run through each pixel in the mask. If over RED_THRESHOLD is white
+    # then we have detected a pylon
+    red_pixels = 0
     for x in range(0, red_mask.shape[0]):
         for y in range(0, red_mask.shape[1]):
             if red_mask[x, y] != 0:
-                return True
+                red_pixels += 1
```
coledie: I believe we can speed this up by using np.sum on the red_mask. Assuming there is only positive values,
ie 
```python
import timeit
import numpy as np


def iteration(img):
    red_pixels = 0
    for x in range(0, img.shape[0]):
        for y in range(0, img.shape[1]):
            if img[x, y] != 0:
                red_pixels += 1
    return red_pixels


def matrix(img):
    return np.sum(img)


if __name__ == '__main__':
    image = np.zeros((720, 1080))
    # image = np.random.uniform(0, 1, (720, 1080))

    t_iter = timeit.timeit(lambda: iteration(image), number=1)
    t_matrix = timeit.timeit(lambda: matrix(image), number=1)

    print(t_iter, t_matrix)  # -> 0.361, 0.001
```

coledie (changes requested):

coledie (approved):

PR #121, opened 2020-02-29
Save depth images as pickle object to preserve information.

coledie:
A pickle object or numpy binary seem to be the only file types to save depth image as, potentially a csv file could work since depth is 2d but I think we should use something that will generalize to higher number channels. I have tried all of the file types that opencv supports but they are all capped at integers of 255 and floats of 1, where the depth images are ~6k+.
I have validated this with a realsense camera. The airsim capture is untested.


ClayMav (approved):

ClayMav:
codacy though

PR #122, opened 2020-02-29
Implement basic algorithm time and boolean accuracy benchmarks.

### LINKED ISSUE #100: Implement algorithm benchmarks.
Benchmarks measuring...
1. Algorithm run time for images of varying resolution / noise
2. Percent accuracy per image class/type - per class confusion matrix.
3. Log list of files failed on so we can see how changes in algorithm affect performance.

### LINKED ISSUE #101: Make benchmarks operate off csv files w/in dataset.
Use csv data to...
1. separate images into classes so that we can tune algorithms to problem areas.
2. Give more information on image collection.
3. Provide more information on image ie module orientation.
This should coincide with the annotation folder but not carry the same information.

coledie:
There is room for many improvements to make this more useful, but as it stands can benchmark algorithm runtime well and accuracy of some functions.


ClayMav (changes requested):
codacy

ClayMav (approved):

PR #141, opened 2020-03-04
Feature/text

### LINKED ISSUE #108: Only return BoundingBoxes with text.

### LINKED ISSUE #127: Rename boat to text.
Convert all usage of boat to text, because we will store mast algorithms in separate directory.
* readme
* unit tests
* benchmarks
* folder
* google drive

nairod2000:
put the text detector into a class and changed all instances of boat to text and changed the algorithm such that it only returns boxes when it has text that we want


vision/tools/blob_annotator/main.py
```diff
@@ -237,7 +237,7 @@ def on_mouse_event(self, event, x, y, flags, params):
 
 
 if __name__ == '__main__':
-    DATASET_PATH = os.path.join("..", "..", "vision_images", "obstacle")
+    DATASET_PATH = os.path.join("../../vision_images/boat/")
```
coledie: This should be text. Make sure to update the folder name of the dataset in the google drive.

coledie (changes requested):
Fix the codacy errors too.

coledie (approved):
Text detection.

PR #148, opened 2020-03-04
Refactor pipeline, respond according to state.

### LINKED ISSUE #91: Way to easily change the type of camera used in pipeline.
Note, different camera objects have different parameters that need to be accounted for somehow. A config file is OK, but more robust, cleaner solutions are encouraged.

### LINKED ISSUE #102: Pipeline change flow of information relative to game state.
Different algorithms need to run based on the current phase of the game. The phase of the game will be given to us by flight as the state.

coledie:


vision/pipeline.py
```diff
@@ -3,62 +3,114 @@
...
+        return state
+
+    def call_me(self):
+        """
+        Alex, call this function - not run.
+        """
+        prev_state = 'start'
+
+        for _ in range(100):
+            prev_state = pipeline.run(prev_state)
```
ClayMav: Would prefer if this was not part of the class, and created an instance of the class and called its run function. You also have a magic number here, 100 calls? Why not just go forever

ClayMav (changes requested):

vision/pipeline.py
coledie: Good idea, the reason it does not go forever is because the bag file reader will just endlessly iterate over videos. It is nice to have it terminate automatically.

vision/pipeline.py
```diff
@@ -3,62 +3,115 @@
...
+        ##
+        for bbox in bboxes:
+            self.vision_communication.put(bbox, self.PUT_TIMEOUT)
+
+        # from vision.common.blob_plotter import plot_blobs
+        # plot_blobs(self.obstacle_finder.keypoints, color_image)
+
+        return state
+
+def init_pipeline(vision_comm, flight_comm, video, runtime=100):
```
ClayMav: just call it vision please so its consistent

ClayMav (changes requested):

vision/pipeline.py
```diff
@@ -3,62 +3,115 @@
...
+        ##
+        for bbox in bboxes:
+            self.vision_communication.put(bbox, self.PUT_TIMEOUT)
+
+        # from vision.common.blob_plotter import plot_blobs
+        # plot_blobs(self.obstacle_finder.keypoints, color_image)
+
+        return state
+
+def init_vision(vision_comm, flight_comm, video, runtime=100):
```
ClayMav, line 88:
```suggestion
def vision(vision_comm, flight_comm, video, runtime=100):
```
Also not sure why there is a runtime, we won't know the runtime

ClayMav: And how are we going to pass you the video? We don't have it in run.py

ClayMav (changes requested):

vision/pipeline.py
coledie: We dont want to name a function the same as a folder, it can be ambiguous to the interpreter.

coledie: I told alex how to create a camera object.

ClayMav: You will be enabling and disabling multiple cameras throughout, it’s gonna be visions responsibility to make those objects

ClayMav: I agree it’s mildly ambiguous but it’s a one time confusion and makes more sense from the run.py perspective

coledie: I will add that to the issue list. Its more like you cant have two things with the same name loaded ie
```python
import numpy

def numpy():
    pass

numpy.ndarray # -> error
```
So naming the function the same as the folder could limit us in the future or cause very weird bugs.

ClayMav: Sure but that won’t apply here as we will only be importing this function outside of vision and you won’t ever have to import that function from inside vision

coledie: It won't apply here but it's not good style, in my opinion.

ClayMav:
Is the codacy error valid? If not, I am good to approve.

coledie:
> Is the codacy error valid? If not, I am good to approve.

No, the error is for a mock function in the unit test.

ClayMav (approved):

PR #151, opened 2020-03-06
Improve blob annotator.

### LINKED ISSUE #48: Improve blob annotator
Improve code quality of blob annotator + Integration w/ unit test(test_blobbing).

coledie:
This was done my peter and jack - not me.


vision/tools/blob_annotator/main.py
```diff
@@ -22,6 +22,7 @@ class PascalVocAnnotator(object):
     CONFIG_PATH = 'config.json'
     SUPPORTED_FILE_EXTENSIONS = ('.jpg', '.png')
 
+
     def __init__(self, path_to_image_folder=TEST_DIRECTORY):
         ## Read config
```
fallscameron01, line 29: Functions are lacking docstrings throughout these files.

fallscameron01 (changes requested):

vision/tools/blob_annotator/main.py
coledie: I agree. All of the classes and files have docstrings though and I think that is sufficient for now.

PR #152, opened 2020-03-06
Linear Actuator Files

### LINKED ISSUE #78: Class to use linear actuator over serial
## Description
For placing the module we will be using a linear actuator (similar to a piston in Minecraft). We can control it using a serial connection, so we need to create a class with accessors like extend (which should go to a set percent extension) as well as being able to read its current status.

https://github.com/RoverRobotics/liblac
https://github.com/DamnedFacts/actuonix-lac

## Requirements
- [ ] Figure out how to interact with LA
- [ ] Compose class with extend, read, and any other useful methods
- [ ] Complete functionality of member methods
- [ ] Do above in own module/files

SercanAkcay:
Closes #78

### DIFF
--- a/flight/utils/Wrapped_LAC.py
+++ b/flight/utils/Wrapped_LAC.py
@@ -58,15 +58,15 @@ stroke = 300 #max length of LAC (mm)
 
 class sLAC:
   def __init__(self):
-    self.piston = LAC(vendorID,productID)
-    self.setupLAC()
-    
     #Ensures the LAC will not hit mechanical stops
     if (rPosition <= rMechStop):
       raise Exception("Retract is set to the mechanical stop.")
     elif (ePosition >= eMechStop):
       raise Exception("Extend is set to the mechanical stop.")
-  
+    
+    self.piston = LAC(vendorID,productID)
+    self.setupLAC()
+    
   #Automatically sets up the LAC
   def setupLAC(self):
     
--- a/flight/utils/Wrapped_LAC.py
+++ b/flight/utils/Wrapped_LAC.py
@@ -1,0 +1,137 @@
+#!/usr/bin/python
+import asyncio
+from lac import *
+from mavsdk import System
+
+'''
+Given:
+  Distance = value/1024 * stroke
+  Stroke is max extension length
+  All values in mm
+  0 and 1023 are the mechanical stops **BAD**
+  Max Speed = 46 mm/s
+  Stroke = 300mm
+  
+Notes:
+  DO NOT set the rLimit or eLimit to their respective mechanical stops
+    (Unless you want to possibly break the Linear Actuator)
+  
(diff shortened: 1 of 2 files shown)


flight/utils/Wrapped_LAC.py
```diff
@@ -0,0 +1,137 @@
...
+    self.piston.set_speed(maxSpeed)
+    print("Piston set to max speed")
+    
+    #Ensures the LAC will not hit mechanical stops
+    if (rPosition <= rMechStop):
+      return False
+    elif (ePosition >= eMechStop):
+      return False
+    else:
+      return True
```
ClayMav: Move the checks for mechstop to the init function, and return nothing from this function

flight/utils/Wrapped_LAC.py
```diff
@@ -0,0 +1,137 @@
+#!/usr/bin/python
+import asyncio
+from lac import *
```
ClayMav: don't use `import *` just import each of the individual elements used

flight/utils/Wrapped_LAC.py
```diff
@@ -0,0 +1,137 @@
+#!/usr/bin/python
+import asyncio
```
ClayMav: is this used

flight/utils/Wrapped_LAC.py
```diff
@@ -0,0 +1,137 @@
+#!/usr/bin/python
```
ClayMav: this should never be called so remove the shebang

flight/utils/Wrapped_LAC.py
```diff
@@ -0,0 +1,137 @@
...
+import asyncio
+from lac import *
+from mavsdk import System
```
ClayMav: Never used

ClayMav (changes requested):
Can't approve until i run it on my machine to test it.

SercanAkcay:
Got it I’ll make the changes

flight/utils/Wrapped_LAC.py
```diff
@@ -0,0 +1,135 @@
...
+class sLAC:
+  def __init__(self):
+    self.piston = LAC(vendorID,productID)
+    self.setupLAC()
+    
+    #Ensures the LAC will not hit mechanical stops
+    if (rPosition <= rMechStop):
+      raise Exception("Retract is set to the mechanical stop.")
+    elif (ePosition >= eMechStop):
+      raise Exception("Extend is set to the mechanical stop.")
```
ClayMav, line 65: move these above everything else in the function

ClayMav (changes requested):

ClayMav (approved):
codacy has a few issues you should fix but I am satisfied once she is

PR #156, opened 2020-03-06
Add logging both to file and console during flight

### LINKED ISSUE #133: Flight Logging to File
## Description of Problem

When we crashed, I was outside of wifi range, so I have no idea where the program was at when the failure occurred, so adding python logging to console and a file will give us viewable logs. https://realpython.com/python-logging/#basic-configurations

## Requirements

- [x] Log to file and console
- [ ] Add logs to all important actions
- [ ] Log all states when entering and transitioning
- [x] Test that errors are begin correctly logged and captured
- [ ] Guarantee that log is being saved to file even when program fails to finish executing -- like when the python interpreter is SIGKILL'd

ClayMav:
By switching from print statements to logging, we have fine control of where our output goes, and we can now log to a file and to console at the same time, allowing for log retrieval in the case of an incident
Closes #133

### DIFF
--- a/flight/flight.py
+++ b/flight/flight.py
@@ -7,8 +7,13 @@ import mavsdk as sdk
 
 from flight.states import STATES, State
 
+SIM_ADDR: str = "udp://:14540"  # Address to connect to the simulator
+CONTROLLER_ADDR: str = "serial:///dev/ttyUSB0"  # Address to connect to a pixhawk board
+
 
 class DroneNotFoundError(Exception):
+    """Exception for when the drone doesn't connect"""
+
     pass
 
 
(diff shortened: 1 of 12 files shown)


flight/flight.py
```diff
@@ -1,52 +1,60 @@
...
         self.drone: System = drone
 
     async def run(self) -> None:
-        """ 
+        """
         Runs the state machine by infinitely replacing current_state until a
         state return None.
         """
         while self.current_state:
             self.current_state: State = await self.current_state.run(self.drone)
```
coledie, line 48: If the while loop ends, would it be good to ensure the drone is landed before just stopping the state machine?

coledie (changes requested):

flight/flight.py
ClayMav: It's not really the responsibility of the state machine and also out of scope for this PR

coledie (approved):

PR #157, opened 2020-03-06
Configuration and GPS Offset for Laps

### LINKED ISSUE #131: Flight configuration File
## Description of Problem

We have several magic numbers and constants in our code so far that should be not only documented but configurable, so to remedy this, we should allow for file-based configuration.

## Requirements

- [ ] Python file containing in some capacity:
    - [ ] Pylon coordinates in a simple latlon data structure
    - [ ] Max speed
    - [ ] Takeoff altitude (basically mission altitude)
    - [ ] Distance thresholds to be used for GPS movements
    - [ ] Any other values you can think of

### LINKED ISSUE #135: Flight Speed and Altitude Limiting
## Description of Problem

During the flight test, the drone had a max speed of almost 50 mph, which was unsafe. Set up maximum speed and altitude functionality to the flight code.

## Requirements

- [ ] Configure the drone object to have a max speed and takeoff altitude 
- [ ] Put in safeguards against altitudes significantly higher than the specified

### LINKED ISSUE #139: GPS Coordinate Offset function
## Description of Problem

For the pylons, we need to be some distance from the actual gps point, so we need a function that when given a gps point (lat, lon) and an angle (where west is 0, north is 90 etc.) spits out a new gps point at that location.

https://pypi.org/project/LatLon/

ewad3:
Closes #131, Closes #135


flight/states/state.py
```diff
@@ -40,6 +40,7 @@ class State:
         drone : System
             The drone system; used for flight operations.
         """
+        await drone.action.arm()
```
ClayMav:
```suggestion
```

flight/states/early_laps.py
```diff
@@ -25,14 +21,15 @@ class EarlyLaps:
...
         """Moves the drone to the first pylon, then begins the 8 laps"""
+        const=Constant()
         print("-- Run Laps")
-        pos_wait = asyncio.ensure_future(self.wait_pos(drone, lat1, lon1))
+        pos_wait = asyncio.ensure_future(self.wait_pos(drone, const.pylon1))
         await pos_wait
         pos_wait.cancel()
         del pos_wait
-        async for i in arange(8):
+        async for i in arange(1):
```
ClayMav:
```suggestion
        async for i in arange(3):
```
at least do a few

flight/states/early_laps.py
```diff
@@ -25,14 +21,15 @@ class EarlyLaps:
 
     async def run(self, drone):
         """Moves the drone to the first pylon, then begins the 8 laps"""
+        const=Constant()
```
ClayMav: should not have been called const, it should be config

ClayMav (changes requested):
Really would rather it not be a class, its better to just import the module (whole file) and then just reference it like `config.config_param(drone)`

In addition I think youve done the submodule wrong, and committed all this other repositories code

ClayMav:
Need to get code changes off of the testing platform before this can be amended and merged

flight/states/early_laps.py
ClayMav: make this a configuration variable

flight/config.py
```diff
@@ -0,0 +1,87 @@
...
+pylon2: LatLon = LatLon(lat2, lon2)
+
+OFFSET: float = .005 #km
+DEG_OFFSET: int = 90 #deg
+
+NUM_LAPS: int = 2
+
+
+
+async def config_param(drone: System):
```
ClayMav:
```suggestion
async def config_params(drone: System):
```

flight/config.py
```diff
@@ -0,0 +1,87 @@
...
+    await drone.param.set_float_param('MIS_TAKEOFF_ALT', TAKEOFF_ALT)
+    await drone.param.set_float_param('MPC_CRUISE_90', MAX_SPEED)
+    await drone.param.set_float_param('MPC_XY_VEL_MAX', MAX_SPEED)
+    await drone.param.set_float_param('MPC_XY_CRUISE', MAX_SPEED)
+    await drone.action.set_maximum_speed(MAX_SPEED)
```
ClayMav, line 47: Comments for the functions of these

flight/config.py
```diff
@@ -0,0 +1,87 @@
...
+    await drone.param.set_int_param('COM_OBL_RC_ACT', 5)
+
+    #Set RC loss failsafe mode HOLD
+    await drone.param.set_int_param('NAV_RCL_ACT', 1)
+
+    await drone.param.set_float_param('LNDMC_XY_VEL_MAX', .5)
+    await drone.param.set_float_param('LNDMC_FFALL_THR', 3)
+    await drone.param.set_float_param('LNDMC_FFALL_TTRI', .15)
+    await drone.param.set_float_param('LNDMC_ALT_MAX', MAX_ALT)
+    await drone.param.set_float_param('LNDMC_LOW_T_THR', .2)
```
ClayMav, line 62: Comments for the functions of these

flight/config.py
```diff
@@ -0,0 +1,87 @@
...
+    
+
+    return
```
ClayMav:
```suggestion
```

flight/config.py
```diff
@@ -0,0 +1,87 @@
...
+#LNDMC_XY_VEL_MAX  m/s 1.5
+#LNDMC_Z_VEL_MAX 
+#LNDMC_LOW_T_THR Low throttle detection threshold .3
+#MAV_0_MODE
+#MAV_0_RATE 
+
+#EKF2
+#Local position Estimator
+#SD Logging
+#Landing target Estimator
```
ClayMav: What is all this

flight/flight.py
```diff
@@ -91,6 +92,8 @@ def flight(comm, sim: bool, log_queue, worker_configurer) -> None:
     """Creates drone object and passes it to start_flight"""
     try:
         drone: System = await init_drone(sim)
+        # config drone param's
+        await config.config_param(drone)
```
ClayMav: This should go in `init_drone`

flight/flight.py
```diff
@@ -133,7 +136,7 @@ def flight(comm, sim: bool, log_queue, worker_configurer) -> None:
             await drone.offboard.set_velocity_ned(sdk.VelocityNedYaw(0, 0, 0, 0))
             await drone.offboard.set_velocity_body(sdk.VelocityBodyYawspeed(0, 0, 0, 0))
 
-            await asyncio.sleep(1)
+            await asyncio.sleep(2)
```
ClayMav: Can you make a config var for the sleep time for thinking

flight/flight.py
```diff
@@ -142,7 +145,7 @@ def flight(comm, sim: bool, log_queue, worker_configurer) -> None:
                     "Stopping offboard mode failed with error code: %s", str(error)
                 )
                 # Worried about what happens here
-            await asyncio.sleep(1)
+            await asyncio.sleep(2)
```
ClayMav: and config var here

flight/states/early_laps.py
```diff
@@ -27,81 +23,75 @@ class EarlyLaps:
...
-            try:  # deturman what velocity should go at
-                dx = math.copysign(35 * math.cos(math.atan(y / x)), x)
-                dy = math.copysign(35 * math.sin(math.atan(y / x)), y)
+            current = LatLon(lat, lon)  # you are here
+
+            if count==0:
+                #offset pylon
+                deg_to_pylon = current.heading_initial(pylon)
+                offset_point = pylon.offset(deg_to_pylon+config.DEG_OFFSET, config.OFFSET)
+                logging.warning(offset_point.to_string('d% %m% %S% %H'))# you are here
```
ClayMav:
```suggestion
                logging.debug(offset_point.to_string('d% %m% %S% %H'))# you are here
```

flight/states/early_laps.py
```diff
@@ -27,81 +23,75 @@ class EarlyLaps:
...
-                    35 * math.sin(math.asin(y / (math.sqrt((x ** 2) + (y ** 2))))), y
+                    config.MAX_SPEED * math.sin(math.asin(y / (math.sqrt((x ** 2) + (y ** 2))))), y
                 )
 
             await drone.offboard.set_velocity_ned(sdk.VelocityNedYaw(dy, dx, alt, deg))
 
-            if abs(x) <= 10 and abs(y) <= 10:
+            logging.debug("CHECKING CLOSENESS")
+            logging.debug("x :: %f <= %f", x, reference_x*config.POINT_PERCENT_ACCURACY)
+            logging.debug("y :: %f <= %f", y, reference_y*config.POINT_PERCENT_ACCURACY)
```
ClayMav:
```suggestion
```

flight/states/early_laps.py
```diff
@@ -112,8 +102,12 @@ class EarlyLaps:
                 temp = (current + 180) % 360
 
             await drone.offboard.set_velocity_body(
-                sdk.VelocityBodyYawspeed(60, -55, 0.25, -90)
+                sdk.VelocityBodyYawspeed(5, -3, -0.1, -60)
             )
-            if current == temp:
+            await asyncio.sleep(1)
```
ClayMav: would also want to add that config var here

flight/states/early_laps.py
```diff
@@ -112,8 +102,12 @@ class EarlyLaps:
...
+            val = abs(current - temp)
+            # Need to add case so that it can overshoot
```
ClayMav:
```suggestion
            # TODO: Add case so that it can overshoot the point and still complete
```

flight/states/early_laps.py
```diff
@@ -112,8 +102,12 @@ class EarlyLaps:
...
+            if val < 10:
+                logging.info("Finished Turn")
```
ClayMav:
```suggestion
                logging.debug("Finished Turn")
```

flight/states/land.py
```diff
@@ -24,7 +24,7 @@ class Land(State):
         await drone.offboard.set_velocity_ned(sdk.VelocityNedYaw(0, 0, 0, 0))
         await drone.offboard.set_velocity_body(sdk.VelocityBodyYawspeed(0, 0, 0, 0))
 
-        await asyncio.sleep(1)
+        await asyncio.sleep(2)
```
ClayMav: Config var here too

ClayMav (changes requested):

ClayMav:
So you are still pushing the latlon code which you shouldn't be so delete that directory and add the submodules. Also, Codacy is not happy so work on that.

ClayMav (approved):

ClayMav:
Needs to be rebased but I will do it, its just a hard one

PR #174, opened 2020-03-20
Feature/manual takeoff

### LINKED ISSUE #168: Manual Takeoff function
## Description of Problem

This function would ditch the `drone.action.takeoff()` function and instead utilize offboard to increase velocity upwards and in the direction of the first pylon.

acbk2b:
Closes #168


flight/states/takeoff.py
```diff
@@ -47,3 +50,60 @@ class Takeoff(State):
             altitude: float = round(position.relative_altitude_m, 2)
             if altitude >= 2:
                 return True
+
+    async def takeoff(self, drone: System) -> None:
```
ClayMav: Specify in a docstring that this function assumes the drone is armed

flight/states/takeoff.py
```diff
@@ -47,3 +50,60 @@ class Takeoff(State):
...
+
+        logging.info("Taking off towards first pylon")
+
+        up_speed: float = -2
```
ClayMav: This should be a file constant, not inside the function

flight/states/takeoff.py
```diff
@@ -47,3 +50,60 @@ class Takeoff(State):
...
+                dy = math.copysign(35 * math.sin(math.asin(y / (math.sqrt((x ** 2) + (y ** 2))))), y)
+
+            # Start the drone pointing in the direction of the first pylon
+            await drone.offboard.set_velocity_ned(
+                sdk.VelocityNedYaw(dy, dx, up_speed, deg)
+            )
+
+            # Loops until the desired altitude has been attained,
+            # then sets the upward velocity to 0 and returns
+            if await self.wait_alt(drone) == True:
```
ClayMav: if wait_alt doesnt return True, then the drone will be in pieces, so dont use it as an if statement, just say
```python
await self.wait_alt(drone)
other_code_that_happens_after()
```

flight/states/takeoff.py
```diff
@@ -47,3 +50,60 @@ class Takeoff(State):
...
+                # If the drone has reached the desired altitude
+                # Leave it moving towards the first pylon,
+                # but set the upwared velocity to 0, so it stops going up
+                await drone.offboard.set_velocity_ned(
+                    sdk.VelocityNedYaw(dy, dx, 0, deg)
+                )
+                return True
```
ClayMav: If it can't return False, just have it return

flight/states/takeoff.py
ClayMav: Also its an int

ClayMav (changes requested):
A few things:
- I think you should have an optional parameter to specify a tuple of (lat, lon) to go towards, so that we can still use takeoff as a simple vertical takeoff if that is not specified.
- The function takeoff should be in a different file, likely within `flight/utils`, so that it can be used other places if required.
- takeoff should also encapsulate all of the setting of set points and starting offboard
- I think you should rebase onto feature/gps_offset so you can use the LatLon library to simplify your math and logic
- I'm going to start enforcing type hints more closely, so make sure to specify function return types

flight/states/early_laps.py
```diff
@@ -45,63 +46,41 @@ class EarlyLaps:
 
     async def wait_pos(self, drone, goal_lat, goal_lon):
         """Goes to a position"""
```
ClayMav: wait_pos is still in this file

flight/states/early_laps.py
```diff
@@ -45,63 +46,41 @@ class EarlyLaps:
...
+            # Added this as a fail safe.  If Ctrl+C is pressed and the exception
+            # is not caught in run.py for whatever reason, set all velocities to 0
+            # and land the drone at whatver the current position is
+            # NOTE: should update to land at specific point once manual land is implemented
+            except KeyboardInterrupt:
+                drone.offboard.set_velocity_ned(sdk.VelocityNedYaw(0.0,0.0,0.0,0.0))
+                drone.action.land()
 
     async def wait_turn(self, drone):
         """Completes a full turn"""
```
ClayMav: wait turn is still in here

flight/utils/drone_takeoff.py
```diff
@@ -0,0 +1,137 @@
...
+up_speed: int = -2
+
+async def wait_alt(drone: System) -> bool:
+    """Checks to see if the drone is near the target altitude"""
+    async for position in drone.telemetry.position():
+        altitude: float = round(position.relative_altitude_m, 2)
+        if altitude >= 2:
+            return True
+
+async def getPosition(drone: System, target_lat: float, target_lon: float):
```
ClayMav:
```suggestion
async def get_distance(drone: System, target_lat: float, target_lon: float):
```

flight/utils/drone_takeoff.py
```diff
@@ -0,0 +1,137 @@
...
+        curr_lat = round(gps.latitude_deg, 8)
+        curr_lon = round(gps.longitude_deg, 8)
+
+        x = ((target_lon - curr_lon)* 40000 * math.cos((target_lat + curr_lat) * math.pi / 360) / 360) * 1000
+
+        y = ((target_lat - curr_lat) * 40000 / 360) * 1000
+
+        return (x, y, altitude)
+
+async def getVelocity(drone: System, target_lat: float, target_lon: float):
```
ClayMav:
```suggestion
async def get_velocity(drone: System, target_lat: float, target_lon: float):
```

ClayMav (changes requested):
Very few things are wrong with this in its current state but itll get effed when we merge the gps_offset stuff

ClayMav (commented):
https://stackoverflow.com/questions/46497578/while-loop-blocks-asyncio-tasks

ewad3:
Outdated

PR #175, opened 2020-03-20
Update vision queue with tuple (time, bboxes).

### LINKED ISSUE #173: Update vision queue as priority queue.

coledie:
Change how the vision queue is updated, simulate a priority queue to our needs.

### DIFF
--- a/vision/pipeline.py
+++ b/vision/pipeline.py
@@ -10,6 +10,7 @@ gparent_dir = os.path.dirname(parent_dir)
 ggparent_dir = os.path.dirname(gparent_dir)
 sys.path += [parent_dir, gparent_dir, ggparent_dir]
 
+import datetime
 import json
 from multiprocessing import Queue
 from queue import Empty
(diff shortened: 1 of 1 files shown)


vision/pipeline.py
```diff
@@ -77,8 +81,7 @@ def run(self, prev_state):
             pass  # raise AttributeError(f"Unrecognized state: {state}")
 
         ##
-        for bbox in bboxes:
-            self.vision_communication.put(bbox, self.PUT_TIMEOUT)
+        self.vision_communication.put((self.time, bboxes), self.PUT_TIMEOUT)
```
ClayMav: Just send the current time in datetime format, then I can prune old ones

ClayMav (changes requested):

vision/pipeline.py
coledie: Is there any specific function I can use that would work best?

ClayMav: datetime.datetime() https://docs.python.org/3/library/datetime.html

coledie: datetime.datetime.now() it is

ClayMav (approved):

PR #181, opened 2020-04-05
Added module_orientation file

### LINKED ISSUE #21: Calculate Module Orientation
Determine the orientation of the module, all 3 axes. Input will be the approximate center of the module, the color and depth images.

Andrew-Gibson42:


vision/module/module_orientation.py
```diff
@@ -0,0 +1,80 @@
+"""
+module_orientation will calculate the orientation of the module in degrees
+and return them as a tuple (x tilt, y tilt) using a derivative through the x- and y- axes
+"""
+import numpy as np
+import argparse
+from region_of_interest import region_of_interest
```
coledie: Lets move these two imports into the if __name__ == '__main__' so they are not confused as necessary for the whole program.

vision/module/module_orientation.py
```diff
@@ -0,0 +1,80 @@
...
+        # x_diffs = x_diffs + np.mean(np.diff(row))
+
+    x_avg_diff = x_diffs / num_rows
+
+    x_tilt = np.degrees(np.arctan(x_avg_diff))
+
+    np.transpose(roi)
+
+    y_diffs = 0
+    for col in roi:
```
coledie: I think it would be better to not modify roi and use the transposed version in the for loop.
```
for col in roi.T:
```

vision/module/module_orientation.py
```diff
@@ -0,0 +1,80 @@
...
+    num_rows, num_cols = roi.shape
+    print("rows: " + str(num_rows))
+    print("cols: " + str(num_cols))
+
+    x_diffs = 0
+    for row in roi:
+        x_diffs = x_diffs + (row[num_cols-1] / 1000 - row[0] / 1000)
+        # x_diffs = x_diffs + np.mean(np.diff(row))
+
+    x_avg_diff = x_diffs / num_rows
```
coledie: This is isometric to
```python
x_avg_diff = np.mean(roi[:, -1] - roi[:, 0]) / 1000
```
Yet this version avoids for loops, so it will be faster with numpy.

coledie: This also applies to the y version.

coledie (changes requested):
Make sure to fix the trailing whitespace errors suggested by codacy.

fallscameron01 (changes requested):
Make sure to fix the issues suggested by Codacy.

vision/module/module_orientation.py
```diff
@@ -0,0 +1,58 @@
...
+
+def get_module_orientation(roi):
+    """
+    Finds the orientation of the module in degrees
+
+    Parameters
+    ----------
+    roi: numpy array
+        module region of interest calculated by region_of_interest
+    """
```
fallscameron01, line 21: Adding a return type to documentation helps us look back on the code later.

fallscameron01 (dismissed):

coledie (approved):

PR #183, opened 2020-04-06
Feature/module_bounding

### LINKED ISSUE #154: Pull module center, nearness and orientation into single bounding box.

fallscameron01:


vision/pipeline.py
```diff
@@ -76,6 +85,21 @@ def run(self, prev_state):
         else:
             pass  # raise AttributeError(f"Unrecognized state: {state}")
 
+        if state == 'module_detection':
```
coledie: Lets delete the previous else: pass, and change this if to and elif
ie
```python
elif state == 'module_detection':
```

vision/module/module_bounding.py
```diff
@@ -0,0 +1,46 @@
...
+"""
+
+import numpy as np
+
+MODULE_HEIGHT = 76.2  # mm
+MODULE_WIDTH = 50.8  # mm
+VERTICAL_FOV = 57  # degrees
+HORIZONTAL_FOV = 86  # degrees
+
+PADDING_CONSTANT = .85
```
coledie: The region of interest uses this variable name and we will want this value to be different than that one. Its highly likely we will want this padding constant to be an overestimate while we will want the roi to be an underestimate.
```python
BBOX_PADDING_CONSTANT = 1.15
```

vision/module/module_bounding.py
```diff
@@ -0,0 +1,46 @@
...
+
+def getModuleBounds(color, center, depth):
+    """
+    getModuleBounds will find four coordinates within the module that will be used to create a BoundingBox.
+
+    Parameters
+    ----------
+    color - ndarray - the color image
```
coledie: For clarity, lets just pass the dimensions of the color image instead of the whole thing.

coledie (changes requested):
This is good stuff - pull develop into this branch and resolve the conflicts too.

coledie (approved):

PR #185, opened 2020-09-15
Hotfix/summer updates

ewad3:
MAVSKD made changes over the summer causing code to fail. This will fix that and do some general code cleanup with pre-commit.


pieperm (approved):
Looking good!

PR #186, opened 2020-09-18
Checks for piston error

JakeLockey:
Catches the exception thrown when failing to connect to the linear actuator, and logs an error message "Failed to connect to the piston". Also added an if statement checking if a linear actuator is connected before each command with the linear actuator.


ewad3 (approved):
Looks solid.

PR #191, opened 2020-10-14
ModuleInFrame modification

### LINKED ISSUE #94: in_frame able to handle background noise more effectively.
The algorithm does not need to handle every picture, the images with a large number of circles in the background are probably irrelevant.
In order to fix the issues we can...
1a. Tune hough circles params to only notice circles on module.
1b. Better preprocess the image to reduce probability of detecting circles in background
2. Tune parameters on parallel line counting / generating to ignore extraneous circles better.
3. Add extra heuristic to only count lines/circles w/in regions of interest.
3. Add an extra metric to reduce false positives / true negatives.
4. Replace current algorithm entirely.

Andrew-Gibson42:
Changed param1 (50 to 75), param2 (28 to 24) and typed ModuleInFrame.

Minimal improvement, but this is the only found combination of parameters that produces a strict improvement for images with the module and without.

One main thing I learned, though, is that the program more or less won't find the module in the frame incorrectly unless we're deliberately trying to trick it.

### DIFF
--- a/vision/module/in_frame.py
+++ b/vision/module/in_frame.py
@@ -8,19 +8,19 @@ import numpy as np
 import argparse
 
 # Constants
-BLUR_SIZE = 5 # Blur kernel size
-BUCKET_MODIFIER = 1 # Changes how many buckets are in the range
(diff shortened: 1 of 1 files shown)


fallscameron01 (approved):
Looks good.

PR #195, opened 2020-10-22
New Manual Takeoff

### LINKED ISSUE #168: Manual Takeoff function

JakeLockey:
Made a function in the takeoff.py file called "takeoff", it runs after the configuration for manual control is done. All it does is set the velocity to be 1.0 upwards, and then uses the wait_alt function to stop at some altitude defined by variable "alt". After the takeoff function completes, it goes to the early_laps file, where the first wait_pos sends the drone to the first pylon.

### DIFF
--- a/flight/states/takeoff.py
+++ b/flight/states/takeoff.py
@@ -17,10 +17,6 @@ class Takeoff(State):
         """Arms and takes off the drone"""
         await self._check_arm_or_arm(drone)  # Arms the drone if not armed
         logging.info("Taking off")
-        # Takeoff command, goes to altitude specified in params
-        await drone.action.takeoff()
-        # waits for altitude to be close to the specified level
-        await self.wait_alt(drone)
 
         # Setting set points for the next 3 lines (used to basically set drone center)
         # (NSm, EWm, DUm, Ydeg)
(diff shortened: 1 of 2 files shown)


ewad3:
The takeoff function should only have drone and not the alt. The variable alt should be a constant in config.py replacing TAKEOFF_ALT with some flying altitude constant and using that new flying altitude constant in the respective math needed.

flight/states/takeoff.py
```diff
@@ -41,15 +37,30 @@ async def run(self, drone: System) -> None:
...
+        except sdk.offboard.OffboardError:
             await drone.action.land()
             return
 
+        await self.takeoff(drone)
+        #Takes off vertically until a desired altitude constant TAKEOFF_ALT
+        #Then moves onto EarlyLaps, were the wait_pos function moves the drone towards the first pylon
         return EarlyLaps()  # Return the next state, RunLaps
 
+    async def takeoff(self, drone: (System)):
```
pieperm: Remove redundant parentheses

flight/states/takeoff.py
```diff
@@ -41,15 +37,30 @@ async def run(self, drone: System) -> None:
...
             await drone.offboard.start()
-        except sdk.OffboardError:
+        except sdk.offboard.OffboardError:
             await drone.action.land()
             return
 
+        await self.takeoff(drone)
+        #Takes off vertically until a desired altitude constant TAKEOFF_ALT
+        #Then moves onto EarlyLaps, were the wait_pos function moves the drone towards the first pylon
         return EarlyLaps()  # Return the next state, RunLaps
```
pieperm: Might as well change this comment to say EarlyLaps instead of RunLaps

pieperm (changes requested):

pieperm (approved):

ewad3 (approved):

PR #197, opened 2020-10-22
Wrapped movement functions

### LINKED ISSUE #137: Functions to wrap movements
## Description of Problem

Currently if any command is rejected by MAVSDK-Python, it raises an exception. We can also run into situations where the drone is not initialized and an action is performed. Checking for that and much more is crucial. Another instance is with land, there are several steps, setting set points, stopping offboard, and landing, we should wrap that functionality. One more thing is when offboard fails to stop, that means we cannot use the built in land mode, so we need to create a new land function that utilizes offboard as a backup.

SercanAkcay:
Wrapped movement functions for the drone and made changes to import/function names.


flight/config.py
```diff
@@ -38,10 +41,12 @@
 
 
 async def config_params(drone: System):
-    await drone.param.set_param_float("MIS_TAKEOFF_ALT", TAKEOFF_ALT)
+    return
```
pieperm: Was keeping this change intentional, or was this for testing purposes?

flight/utils/Wrapped_Movement.py
```diff
@@ -0,0 +1,115 @@
+from flight import config
```
pieperm: Add a short docstring at the top here to describe the purpose of the file

e.g. `"""Provides a set of reusable movement functions for the drone"""`

flight/states/early_laps.py
```diff
@@ -7,7 +7,7 @@
 from flight import config
 from flight.utils.latlon import LatLon
 
-
+from flight.utils.Wrapped_Movement import D_Move
```
pieperm: `Wrapped_Movement.py` could be renamed to `movement_controller.py`. Since this is now in utils, "movement controller" more accurately describe what these functions do, rather than how they were changed.

flight/utils/Wrapped_Movement.py
```diff
@@ -0,0 +1,115 @@
...
+import logging
+import asyncio
+import math
+
+class D_Move:
+    """
+    Params: Drone, Pylon
+    Calculates and uses gps coordinates of the drone to move to the location of the target pylon 
+    """
+    async def moveTo(self, drone: System, pylon: LatLon):
```
pieperm: Function docstrings go beneath the function definition, indented. For example:
```
def __init__(self, initial_state: State, drone: System) -> None:
        """
        The constructor for StateMachine class.
        Parameters:
            current_state (State): State that is currently being executed.
            drone (System): Our drone object; used for all flight functions.
        """
```

flight/utils/Wrapped_Movement.py
```diff
@@ -0,0 +1,115 @@
...
+                logging.debug(offset_point.to_string("d% %m% %S% %H"))  # you are here
+            # distance we have to go in order to get to the offset point
+            dist = current.distance(offset_point)
+            # degrees needed to change to get to offset position
+            deg = current.heading_initial(offset_point)
+
+            #East, West
+            x = dist * math.sin(math.radians(deg)) * 1000  # from km to m
+            # North, South
+            y = dist * math.cos(math.radians(deg)) * 1000  # from km to m
```
pieperm: Add type annotations to `x` and `y` variables

`x: float = ...`
`y: float = ...`

flight/utils/Wrapped_Movement.py
```diff
@@ -0,0 +1,115 @@
...
+                # Creating a new position we need to go to
+                # Distance to the offset point from the pylon
+                offset_point = pylon.offset(
+                    deg_to_pylon + config.DEG_OFFSET, config.OFFSET
+                )
+                logging.debug(offset_point.to_string("d% %m% %S% %H"))  # you are here
+            # distance we have to go in order to get to the offset point
+            dist = current.distance(offset_point)
+            # degrees needed to change to get to offset position
+            deg = current.heading_initial(offset_point)
```
pieperm: Add type annotations to `dist` and `deg` variables

flight/utils/Wrapped_Movement.py
```diff
@@ -0,0 +1,115 @@
...
+            #Only for first run through loop
+            if count == 0:
+                # How many degrees we need to turn in order to look at the pylon
+                # Think of a unit circle
+                deg_to_pylon = current.heading_initial(pylon)
+                # Creating a new position we need to go to
+                # Distance to the offset point from the pylon
+                offset_point = pylon.offset(
+                    deg_to_pylon + config.DEG_OFFSET, config.OFFSET
+                )
```
pieperm: Add type annotation for `offset_point`

flight/utils/Wrapped_Movement.py
```diff
@@ -0,0 +1,115 @@
+from flight import config
+from flight.utils.latlon import LatLon
+from mavsdk import System
+import mavsdk as sdk
+import logging
+import asyncio
+import math
+
+class D_Move:
```
pieperm: Let's change this class name to match the new name of the file:
`class MovementController:`

flight/utils/Wrapped_Movement.py
```diff
@@ -0,0 +1,115 @@
...
+        count = 0
+        async for tel in drone.telemetry.attitude_euler():
+            current = (360 + round(tel.yaw_deg)) % 360
+            if count == 0:
+                temp = (current + 180) % 360
+
+            await drone.offboard.set_velocity_body(
+                sdk.offboard.VelocityBodyYawspeed(5, -3, -0.1, -60)
+            )
+            # await asyncio.sleep(config.FAST_THINK_S)
```
pieperm: Is this line supposed to be commented out? It looks like it was commented before you made changes.

flight/utils/Wrapped_Movement.py
```diff
@@ -0,0 +1,115 @@
...
+
+    """
+    Params: Drone
+    Checks the altitude of the drone to make sure that we are at our target
+    """
+    async def checkAlt(self, drone: System):
+        async for position in drone.telemetry.position():
+            altitude: float = round(position.relative_altitude_m, 2)
+            if altitude >= config.ALT_RANGE_MIN:
+                return True
```
pieperm: This is a PEP8 style thing, but files should end with a black newline

flight/utils/Wrapped_Movement.py
```diff
@@ -0,0 +1,115 @@
...
+            count += 1
+            
+    """
+    Params: Drone
+    Turns the drone 180 degrees around the pylon it is currently at
+    """
+    async def turnTo(self, drone: System):
+        count = 0
+        async for tel in drone.telemetry.attitude_euler():
+            current = (360 + round(tel.yaw_deg)) % 360
```
pieperm: Type annotation: `current: float`

flight/utils/Wrapped_Movement.py
```diff
@@ -0,0 +1,115 @@
...
+            ):
+                return True
+            count += 1
+            
+    """
+    Params: Drone
+    Turns the drone 180 degrees around the pylon it is currently at
+    """
+    async def turnTo(self, drone: System):
+        count = 0
```
pieperm: Type annotation: `count: int`

flight/utils/Wrapped_Movement.py
pieperm: This docstring should include information about the return type -- it looks like there's a `return True` in the function.

pieperm: Also, add a type annotation for the return type:
`async def moveTo(self, drone: System, pylon: LatLon) -> bool:`

flight/utils/Wrapped_Movement.py
```diff
@@ -0,0 +1,115 @@
...
+class D_Move:
+    """
+    Params: Drone, Pylon
+    Calculates and uses gps coordinates of the drone to move to the location of the target pylon 
+    """
+    async def moveTo(self, drone: System, pylon: LatLon):
+        count = 0
+
+        async for gps in drone.telemetry.position():
+            altitude = round(gps.relative_altitude_m, 2)
```
pieperm: Type annotation: `altitude: float`

flight/utils/Wrapped_Movement.py
```diff
@@ -0,0 +1,115 @@
...
+import asyncio
+import math
+
+class D_Move:
+    """
+    Params: Drone, Pylon
+    Calculates and uses gps coordinates of the drone to move to the location of the target pylon 
+    """
+    async def moveTo(self, drone: System, pylon: LatLon):
+        count = 0
```
pieperm: Type annotation: `count: int`

flight/utils/Wrapped_Movement.py
```diff
@@ -0,0 +1,115 @@
...
+                alt = config.ALT_CORRECTION_SPEED  # go down m/s
+            elif altitude <= config.ALT_RANGE_MIN:
+                alt = -config.ALT_CORRECTION_SPEED  # go up m/s
+            else:
+                alt = -0.15  # don't move
+
+            #Configure current position and store it
+            lat = round(gps.latitude_deg, 8)
+            lon = round(gps.longitude_deg, 8)
+            current = LatLon(lat, lon)  # you are here
```
pieperm: Type annotations:
```
lat: float = ...
lon: float = ...
current: LatLon = ...
```

flight/utils/Wrapped_Movement.py
```diff
@@ -0,0 +1,115 @@
...
+
+            #Only for first run through loop
+            if count == 0:
+                # How many degrees we need to turn in order to look at the pylon
+                # Think of a unit circle
+                deg_to_pylon = current.heading_initial(pylon)
```
pieperm: Type annotation for `deg_to_pylon`

flight/utils/Wrapped_Movement.py
```diff
@@ -0,0 +1,115 @@
...
+                dx = math.copysign(
+                    config.MAX_SPEED
+                    * math.cos(math.asin(y / (math.sqrt((x ** 2) + (y ** 2))))),
+                    x
+                )
+                dy = math.copysign(
+                    config.MAX_SPEED
+                    * math.sin(math.asin(y / (math.sqrt((x ** 2) + (y ** 2))))),
+                    y
+                )
```
pieperm: Add type annotations for `dx` and `dy` variables

flight/utils/Wrapped_Movement.py
```diff
@@ -0,0 +1,115 @@
...
+                and abs(y) <= reference_y * config.POINT_PERCENT_ACCURACY
+            ):
+                return True
+            count += 1
+            
+    """
+    Params: Drone
+    Turns the drone 180 degrees around the pylon it is currently at
+    """
+    async def turnTo(self, drone: System):
```
pieperm: Docstring goes beneath the function declaration. Also make sure to add info about the return value.

pieperm: Add return value type annotation:
`async def turnTo(self, drone: System) -> bool:`

flight/utils/Wrapped_Movement.py
pieperm: Let's rename this function from `moveTo` to `move_to` to stick with Python naming conventions.

flight/utils/Wrapped_Movement.py
pieperm: Let's rename this function from `turnTo` to `turn` or `turn_around`. There's no angle parameter that one would expect to be there when instructing to "turn to" something.

flight/utils/Wrapped_Movement.py
```diff
@@ -0,0 +1,115 @@
...
+            # TODO: Add case so that it can overshoot the point and still complete
+            if val < 10:
+                logging.debug("Finished Turn")
+                return True
+            count += 1
+
+    """
+    Params: Drone
+    Checks the altitude of the drone to make sure that we are at our target
+    """
```
pieperm: Docstring goes beneath the function declaration. Also make sure to add info about the return value.

flight/states/early_laps.py
```diff
@@ -16,106 +16,26 @@ async def arange(count):
...
-            await self.wait_turn(drone)  # turn around pylon 2
+            await D_Move.turnTo(self, drone)  # turn around pylon 2
 
             logging.debug("Lap %d: Straight two", i)
-            await self.wait_pos(drone, config.pylon1)  # move to pylon 1
+            await D_Move.moveTo(self, drone, config.pylon1)  # move to pylon 1
 
             logging.debug("Lap %d: Turn two", i)
-            await self.wait_turn(drone)  # turn around pylon 1
+            await D_Move.turnTo(self, drone)  # turn around pylon 1
```
pieperm: You will likely need to update/change some of these names based on suggestions in other comments.

flight/utils/Wrapped_Movement.py
```diff
@@ -0,0 +1,115 @@
...
+    async def checkAlt(self, drone: System):
```
pieperm: Type annotation for return value:
`async def checkAlt(self, drone: System) -> bool:`

pieperm: Let's rename this function from `checkAlt` to `check_altitude`

flight/states/early_laps.py
pieperm: It doesn't make much sense to pass `self` into these functions, since `self` is of type `EarlyLaps` and is not used within the `moveTo` and `turnTo` functions. Currently these functions are being called on the class itself, rather than an instance of the class. To resolve this, all we have to do is instantiate an object before calling these functions:

```
mover: D_Move = D_Move()

class EarlyLaps:
...
```

Which then allows us to call the functions on the object:
`await mover.moveTo(drone, config.pylon1)`
`await mover.turnTo(drone)`
without needing to pass in an extraneous `self` object.

pieperm: Here's an example that's based on the naming changes suggested in this review:
```
movement_controller: MovementController = MovementController()

class EarlyLaps:
...

await movement_controller.move_to(drone, config.pylon1)
await movement_controller.turn(drone)
```

pieperm (changes requested):

flight/config.py
SercanAkcay: testing purposes, would not work on my machine without the change

flight/utils/Wrapped_Movement.py
SercanAkcay: I'll be honest I am not sure

flight/utils/Wrapped_Movement.py
SercanAkcay: What do you mean? Should I add a "\n"?

flight/utils/Wrapped_Movement.py
SercanAkcay: I will figure out why this returns true and where it returns to. As of right now I am not sure.

flight/config.py
pieperm: Let's go ahead and get rid of the return statement, we can always add it back later if it continues causing errors.

flight/utils/Wrapped_Movement.py
pieperm: I think we'll just leave the comment as is for now.

flight/utils/Wrapped_Movement.py
pieperm: All you have to do is hit enter on that last line (after `return True`), and then leave the line blank.

flight/utils/Wrapped_Movement.py
pieperm: It looks like it either returns a `bool` or `None`, but I'm not entirely sure. The function returns true if it gets within range of the target position, but assuming it gets through that for loop (`async for gps in drone.telemetry.position()`) without ever meeting that condition, it looks like it would return `None`

pieperm (dismissed):

ewad3 (approved):

PR #205, opened 2020-11-06
Merge module algorithm parts

### LINKED ISSUE #145: Merge portions of module algorithms to reduce latency & recomputation.

mrouie:
Merged ModuleInFrame into a ModuleLocaiton.isInFrame

### DIFF
--- a/vision/module/in_frame.py
+++ b/vision/module/in_frame.py
@@ -14,6 +14,7 @@ MIN_SLOPES_IN_BUCKET = (
     15  # Minimum number of slopes in a single bucket to identify the module
 )
 MAX_CIRCLES = 100  # Maximum number of cirlces that can be detected in an image before ModuleInFrame fails
+MIN_CIRCLES = 4 # Minimum number of circles needed to perform calculations
 
 
 def ModuleInFrame(color_image: np.ndarray) -> bool:
@@ -57,8 +58,8 @@ def ModuleInFrame(color_image: np.ndarray) -> bool:
         maxRadius=50,
     )
     if (
-        circles is None or circles.shape[1] > MAX_CIRCLES
-    ):  # no circles found or too many circles found
+        np.shape(self.circles) < MIN_CIRCLES or np.shape(self.circles) > MAX_CIRCLES
+    ):  # too little or too many circles found
         return False
 
     circles = np.uint16(circles)
--- a/vision/module/in_frame.py
+++ b/vision/module/in_frame.py
@@ -10,7 +10,9 @@ import argparse
 # Constants
 BLUR_SIZE = 5  # Blur kernel size
 BUCKET_MODIFIER = 1  # Changes how many buckets are in the range
-MIN_SLOPES_IN_BUCKET = 15  # Minimum number of slopes in a single bucket to identify the module
+MIN_SLOPES_IN_BUCKET = (
+    15  # Minimum number of slopes in a single bucket to identify the module
+)
 MAX_CIRCLES = 100  # Maximum number of cirlces that can be detected in an image before ModuleInFrame fails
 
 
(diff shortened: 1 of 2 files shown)


vision/module/location.py
```diff
@@ -34,9 +34,37 @@ def __init__(self):
...
+        # Get slopes and group parallel slopes
+        self._getSlopes()
+        self._groupSlopes()
+
+        return any(self.slope_heights > MIN_SLOPES_IN_BUCKET)
+
     ## Finding Distance to Module
 
-    def getDistance(self):
+    def getDistance(self) -> int:
```
fallscameron01: The getDistance function can be removed from location. A more robust calculation is performed by get_module_depth.py

vision/module/location.py
```diff
@@ -285,7 +313,7 @@ def _increaseBrightness(self, increase):
 
     ## Input Functions
 
-    def setImg(self, color, depth):
+    def setImg(self, color, depth) -> None:
```
fallscameron01: > color, depth

Make sure you type function parameters as well. Should be:
color: np.ndarray, depth: np.ndarray

vision/module/location.py
```diff
@@ -266,7 +294,7 @@ def _filterDepth(self):
         tempDepth = np.dstack((self.depth, self.depth, self.depth))
         self.img = np.where(tempDepth < DEPTH_THRESH, self.img, 0)
 
-    def _increaseBrightness(self, increase):
+    def _increaseBrightness(self, increase) -> None:
```
fallscameron01: Type parameter here as well.

vision/module/location.py
```diff
@@ -49,7 +77,7 @@ def getDistance(self):
 
     ## Finding the Center
 
-    def getCenter(self):
+    def getCenter(self) -> tuple:
```
fallscameron01, line 77: You should add a parameter here that allows us to use previous calculations if in_frame returned true. For example, you could have a bool parameter that defaults to true called recalculate. Then you could check whether recalculate is true before getCenter() runs _getSlopes() and _groupSlopes() again.

fallscameron01 (changes requested):
Looks like you adapted in_frame well. The main thing is that the point of this is to avoid recalculation, so adding a bool parameter like I described to getCenter() would help.

vision/module/location.py
```diff
@@ -34,22 +34,45 @@ def __init__(self):
...
+            15  # Minimum number of slopes per bucket to identify the module
+        )
+        MAX_CIRCLES = 100  # maximum number of circles that are allowed to be detected before in_frame fails
+
+        if self.needsRecalc:
+            # Circle Detection
+            self._circleDetection()
+
+        if (
+            self.circles is None or self.circles.shape[0] > MAX_CIRCLES
```
fallscameron01: > self.circles is None

This should be changed from is None to a constant called MIN_CIRCLES that would be set to 4 since there is also no point to calculations if there is 1, 2, or 3 circles.

fallscameron01 (changes requested):
Just a small change here and the PR good to go.

vision/module/location.py
```diff
@@ -34,22 +34,46 @@ def __init__(self):
...
+        )
+        MAX_CIRCLES = 100  # maximum number of circles that are allowed to be detected before in_frame fails
+        MIN_CIRCLES = 4  # minimum number of circles needed to perform calculations
+
+        if self.needsRecalc:
+            # Circle Detection
+            self._circleDetection()
+
+        if (
+            np.shape(self.circles) < MIN_CIRCLES or np.shape(self.circles) > MAX_CIRCLES
```
fallscameron01: You have to index np.shape(self.circles) since self.circles is a two-dimensional array, otherwise you would be comparing a tuple to an int. should be:

`np.shape(self.circles)[0] < MIN_CIRCLES or np.shape(self.circles)[0] > MAX_CIRCLES`

vision/module/in_frame.py
```diff
@@ -50,11 +47,20 @@ def ModuleInFrame(color_image: np.ndarray) -> bool:
...
+        method=cv2.HOUGH_GRADIENT,
+        dp=1,
+        minDist=8,
+        param1=75,
+        param2=24,
+        minRadius=0,
+        maxRadius=50,
+    )
+    if (
+        np.shape(self.circles) < MIN_CIRCLES or np.shape(self.circles) > MAX_CIRCLES
```
fallscameron01: Same thing here.

fallscameron01 (changes requested):
You need to index np.shape since it will return a 2-tuple for the 2d array self.circles

fallscameron01 (approved):
Looks good.

PR #206, opened 2020-11-18
Added module roll function to module_orientation.py

### LINKED ISSUE #187: Implement roll for orientation.
Implement roll for orientation. May not be needed, depends on how it's implemented in hardware.

Andrew-Gibson42:
get_module_roll has been added to module_orientation.py

To test, either copy the function into module_bounding.py and run from there, or import module_bounding's getModuleBounds into module_orientation.py. Since there is no yet a working center function, to my knowledge, manually input the center/depth indices for the image you want to test. Commented visualization lines have also been added for convenience.

### DIFF
--- a/vision/module/module_orientation.py
+++ b/vision/module/module_orientation.py
@@ -3,6 +3,7 @@ module_orientation will calculate the orientation of the module in degrees
 and return them as a tuple (x tilt, y tilt) using a derivative through the x- and y- axes
 """
 import numpy as np
+import cv2
 
 
 def get_module_orientation(roi):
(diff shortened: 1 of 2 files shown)


vision/module/module_orientation.py
```diff
@@ -22,13 +23,61 @@ def get_module_orientation(roi):
...
     x_tilt = np.degrees(np.arctan(x_avg_diff))
 
-
     y_avg_diff = np.mean(roi.T[:, -1] / 1000 - roi.T[:, 0] / 1000)
     y_tilt = np.degrees(np.arctan(y_avg_diff))
 
     return x_tilt, y_tilt
 
 
+def get_module_roll(enclosing_region):
```
fallscameron01: You need to type the parameter and return type.

vision/module/module_orientation.py
```diff
@@ -22,13 +23,61 @@ def get_module_orientation(roi):
...
+    # contours only work on grey images
```
fallscameron01, line 49: Also, you are missing a function documentation head.

vision/module/module_orientation.py
```diff
@@ -22,13 +23,61 @@ def get_module_orientation(roi):
...
+        box = cv2.boxPoints(rect)
+        box = np.int0(box)
+
+        # visualization
+        # cv2.drawContours(colorImage, [box], 0, (0, 0, 255), 2)
+
+        # if the angle is absurd, it won't be taken
+        #   if the module is truly rolled at over 45 degrees, minAreaRect
+        #   should simply choose a different rectangle or axis
+        if abs(rect[2]) < 45:
```
fallscameron01: This can be done faster using np.where to create a mask.

vision/module/module_orientation.py
```diff
@@ -22,13 +23,61 @@ def get_module_orientation(roi):
...
+            sum_angle += rect[2]
+            divisor += 1
+
+    # averages the angles of the enclosing min area rectangles around the contours
+    roll = sum_angle / divisor
```
fallscameron01: You can use np.mean after np.where masking instead of looping through the rectangles and summing. Looping is much slower than using numpy functions.

vision/module/module_orientation.py
```diff
@@ -22,13 +23,61 @@ def get_module_orientation(roi):
...
+    # use the following line to visualize contours, pair with imshow
+    # colorImage = cv2.drawContours(enclosing_region, contours, -1, (0, 255, 0), 3)
+
+    # simple contours will contain the list of contours with fewer than 8 vertices
+    simple_contours = []
+    for c in contours:
+        approx = cv2.approxPolyDP(c, 0.01 * cv2.arcLength(c, True), True)
+        if len(approx) < 8:
+            simple_contours.append(c)
+            enclosing_region = cv2.drawContours(enclosing_region, c, -1, (0, 255, 0), 3)
```
fallscameron01: Drawing function should be commented out.

fallscameron01 (changes requested):
Good structure. The loop near the bottom can be replaced with numpy functions to improve speed.

vision/module/module_orientation.py
```diff
@@ -22,13 +25,72 @@ def get_module_orientation(roi):
...
+    #     box = np.int0(box)
+    #     cv2.drawContours(colorImage, [box], 0, (0, 0, 255), 2)
+
+    np_rectangles = np.asarray(rectangles)
+    angles = np_rectangles[:, 2]
+
+    # if the angle is absurd, it won't be taken
+    #   if the module is truly rolled at over 45 degrees, minAreaRect
+    #   should simply choose a different rectangle or axis
+    np.where(abs(angles) < 45, angles, angles)
```
fallscameron01: This np.where statement does nothing. It says that if the angles < 45, the value is angles; otherwise, the value is angles. Also, you are not assigning the result and np.where will not change angles.


You should use np.where to create a mask for angles since you need to remove values from angles.
`mask = np.where(abs(angles) < 45)`
This will give you the indexes of the angles less than 45.

vision/module/module_orientation.py
```diff
@@ -22,13 +25,72 @@ def get_module_orientation(roi):
...
+    roll = np.mean(angles)
```
fallscameron01: Then in np.mean, you can index the angles using the mask:
`np.mean(angles[mask])`

fallscameron01 (changes requested):
Need to make some changes to make the mean work correctly.

fallscameron01 (approved):
Looks good.

PR #210, opened 2020-11-20
Make feature/flight_unit_test 

### LINKED ISSUE #209: Flight Unit Test
## Description of Problem

As the code gets more complex the ability to run segments of the code is need more and more.

ewad3:
Also updates poetry lock to allow for vision and flight pipeline.
Closes #209

### DIFF
--- a/flight/config.py
+++ b/flight/config.py
@@ -41,6 +41,8 @@ NUM_LAPS: int = 2
 THINK_FOR_S: float = 2.0
 FAST_THINK_S: float = 1.0
 
+run_states = {"early_laps": False}
+
 
 async def config_params(drone: System):
     await drone.param.set_param_float("MIS_TAKEOFF_ALT", TAKEOFF_ALT)
--- a/flight/states/early_laps.py
+++ b/flight/states/early_laps.py
@@ -7,6 +7,7 @@ import mavsdk as sdk
 from flight import config
 
 from flight.utils.movement_controller import MovementController
+from .state import State
 from .land import Land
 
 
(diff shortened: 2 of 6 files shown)


flight/config.py
```diff
@@ -41,6 +41,8 @@
 THINK_FOR_S: float = 2.0
 FAST_THINK_S: float = 1.0
 
+run_states = {"early_laps": False}
```
pieperm: Since the `ToMast` state is now in `develop`, we can add the option for `"to_mast"` as well.

flight/states/early_laps.py
```diff
@@ -16,27 +17,30 @@ async def arange(count):
...
+                await mover.turn(drone)  # turn around pylon 2
+
+                logging.debug("Lap %d: Straight two", i)
+                await mover.move_to(drone, config.pylon1)  # move to pylon 1
+
+                logging.debug("Lap %d: Turn two", i)
+                await mover.turn(drone)  # turn around pylon 1
+            return Land()
+        else:
+            return Land()
```
pieperm: For this part, we're going to want to transition to `ToMast()` instead of landing.

flight/config.py
pieperm: You'll also have to go into `to_mast.py` and wrap the contents of the `run` function with the `run_states` option, just as you did in `EarlyLaps`

pieperm: It looks like I overlooked changing the string in `__init__.py` from `"mast": ToMast` to `"to_mast": ToMast`, so if you could fix that in this PR, that would be great.

pieperm (changes requested):
A few changes based on the new state `ToMast` that was recently merged into `develop`.

flight/states/early_laps.py
```diff
@@ -21,7 +20,7 @@ class EarlyLaps(State):
 
     async def run(self, drone):
         """Moves the drone to the first pylon, then begins the 8 laps"""
-        if config.run_states["early_laps"]:
+        if config.run_states["earrly_laps"]:
```
pieperm: Was this change intentional?

pieperm (changes requested):

pieperm (approved):
Looks good!

PR #211, opened 2020-11-20
Pipeline update for roll, isInFrame, text detection

fallscameron01:
- include module roll function
- include isInFrame function and testing before further module calculations
- include text detection


pieperm (approved):
Went ahead and approved, but if you have time to add some type annotations, that would be great. If not you can go ahead and merge this and we can worry about adding typing in a future PR.

PR #212, opened 2020-11-20
Feature/updated manual land

### LINKED ISSUE #201: Improve Manual Land function
Write the function to look like the inverse of Manual Takeoff.
Look into using barometer instead of GPS to land the drone.
As we learned form the flight test have it fly to launch positon and not home position because they are not always equal.

cjhr95:
Updated manual landing functions to move drone to initial takeoff location. Also shifted movement functions into movement_controller.py file


ewad3 (approved):
Looks ready to fly.

PR #224, opened 2021-02-12
Track persistent obstacles between frames and update pipeline to utilize ObstacleTracker

### LINKED ISSUE #155: Ensure obstacle present for so many frames before sending to flight.
See #54 sometimes a large, fast movement of the camera causes the image to become very blurry. When this happens its common for many random obstacles to be detected for a frame or two.
The algorithm should probably be more sensitive to horizontal changes vs vertical.

### LINKED ISSUE #202: Track obstacles between frames
Need to be able to track obstacles between frames once they are detected.

mrouie:
This pull request uses ObstacleTracker to track obstacles between frames and updates the pipeline to utilize ObstacleTracker functions

ObstacleTracker uses ObstacleTracker.update() to update the internal buffer of boundingboxes in the current frame with a 10% (radius) movement tolerance per obstacle. 
It also uses ObstacleTracker.getPersistentObstacles() to return a list of BoundingBoxes that contain obstacles that have been persistent for 5 frames.

### DIFF
--- a/vision/obstacle/obstacle_tracker.py
+++ b/vision/obstacle/obstacle_tracker.py
@@ -18,7 +18,7 @@ class Obstacle:
     """
     def __init__(self, bbox: BoundingBox):
         self.bounding_box = bbox
-        self.center = np.mean(bbox)
+        self.center = np.mean(bbox.vertices, axis=0)
         self.frames_persisted = 0
 
 
@@ -27,7 +27,7 @@ class ObstacleTracker:
     Track obstacles between frames.
     """
     def __init__(self):
-        self.MVMT_TOLERANCE = 0.10 # bounding box displacement tolerance per frame
+        self.MVMT_TOLERANCE = 30 # bounding box displacement tolerance per frame (in pixels)
         self.PERSISTENCE_THRESHOLD = 5 # number of frames for an obstacle to be considered consequential
         
         self.obstacles = np.array([]) # buffer that stores current Obstacles in view
@@ -42,8 +42,29 @@ class ObstacleTracker:
             The obstacle of reference
         obj2: Obstacle
             The obstacle in question
+        
+        Returns
+        -------
+        bool
+            The result of whether or not obj1 and obj2 are the same object
         """
-        return obj1.center >= (1-self.MVMT_TOLERANCE)*obj2.center and obj1.center <= (1+self.MVMT_TOLERANCE)*obj2.center
+        isSame = True # presume they're the same obstacle
+
+        # horizontal check
+        if(
+            obj1.center[0] >= (obj2.center[0]+MVMT_TOLERANCE) or # outside of upper bound
+            obj1.center[0] <= (obj2.center[0]-MVMT_TOLERANCE) # less than lower bound
+        ):
+            isSame = False
+
+        # vertical check
+        if(
+            obj1.center[1] >= (obj2.center[1]+MVMT_TOLERANCE) or # outside of upper bound
+            obj1.center[1] <= (obj2.center[1]-MVMT_TOLERANCE) # less than lower bound
+        ):
+            isSame = False
+
+        return isSame
 
     def update(self, new_obstacle_boxes: list) -> None:
         """
@@ -61,7 +82,7 @@ class ObstacleTracker:
         new_obstacles = np.array([])
         
         # transform the list of bounding boxes into an array of Obstacles
-        for i in np.arange(new_obstacle_boxes.size()):
+        for i in np.arange(new_obstacle_boxes.size):
             new_obstacles[i] = Obstacle(new_obstacle_boxes[i])
         
         if(not self.obstacles):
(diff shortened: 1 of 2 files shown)


vision/obstacle/obstacle_tracker.py
```diff
@@ -0,0 +1,95 @@
...
+    def _isSameObstacle(self, obj1: Obstacle, obj2: Obstacle) -> bool:
+        """
+        Returns True if the position of obj1 is within self.MVMT_TOLERANCE of obj2 
+
+        Parameters
+        ----------
+        obj1: Obstacle
+            The obstacle of reference
+        obj2: Obstacle
+            The obstacle in question
```
fallscameron01, line 44: Return in documentation

vision/obstacle/obstacle_tracker.py
```diff
@@ -0,0 +1,95 @@
...
+        """
+        return obj1.center >= (1-self.MVMT_TOLERANCE)*obj2.center and obj1.center <= (1+self.MVMT_TOLERANCE)*obj2.center
```
fallscameron01: Looks a little messy, should put parenthesis around the statements with the and in between:
`(obj1.center >= (1-self.MVMT_TOLERANCE)*obj2.center) and (obj1.center <= (1+self.MVMT_TOLERANCE)*obj2.center)`

vision/obstacle/obstacle_tracker.py
```diff
@@ -0,0 +1,95 @@
...
+        # transform the list of bounding boxes into an array of Obstacles
+        for i in np.arange(new_obstacle_boxes.size()):
+            new_obstacles[i] = Obstacle(new_obstacle_boxes[i])
+        
+        if(not self.obstacles):
+            self.obstacles = new_obstacles
+            return
+
+        # create an array of obstacles that existed in the previous (buffer) frame and replace the buffer with it
+        for i in np.arange(new_obstacles.size()):
```
fallscameron01: `new_obstacles.size()`

Size is actually a variable of an ndarray, not a function. No parenthesis needed.

vision/obstacle/obstacle_tracker.py
```diff
@@ -0,0 +1,95 @@
...
+            if(_isSameObstacle(new_obstacles[i], self.obstacles[i])):
```
fallscameron01: Should be

`self._isSameObstacle(`

since _isSameObstacle is a member function.

vision/obstacle/obstacle_tracker.py
```diff
@@ -0,0 +1,95 @@
...
+    def getPersistentObstacles(self) -> list:
+        """
+        returns a list of bounding boxes that have been persistent for PERSISTENCE_THRESHOLD amount of frames
+
+        Returns
+        -------
+        list[BoundingBox]
+            a list of the persistent obstacle BoundingBoxes
+        """
+        persistent_obstacles = np.array([])
```
fallscameron01: Your return value should be a normal Python list since that is how bounding boxes are stored in the pipeline.

vision/obstacle/obstacle_tracker.py
```diff
@@ -0,0 +1,95 @@
...
+
+        # append only obstacles persistent for PERSISTENCE_THRESHOLD frames
+        for i in np.arange(self.obstacles.size()):
```
fallscameron01: Same issue above, size should not have parenthesis.

vision/obstacle/obstacle_tracker.py
```diff
@@ -0,0 +1,95 @@
...
+    Properties of an obstacle in the frame.
+
+    Parameters
+    ----------
+    bbox: BoundingBox
+        bounding box of obstacle 
+    """
+    def __init__(self, bbox: BoundingBox):
+        self.bounding_box = bbox
+        self.center = np.mean(bbox)
```
fallscameron01: `np.mean(bbox)` will not work because bbox is a user created class. You need to take the mean of bbox.vertices
Also, we want to separate the x and y values, so we should take the mean along the 0th axis.

`np.mean(bbox.vertices, axis=0)` will return [xval, yval] with a datatype of ndarray and datatypes of x and y as floats.

vision/obstacle/obstacle_tracker.py
fallscameron01: Since the center has both an x and a y value, you need to compare x and y separately, then logical and the results from each.

fallscameron01: One more thing with the math here: You are multiplying the center value by the movement tolerance, but you want to be multiplying by the total dimensions of the image. Otherwise, obstacles closer to the top left corner will have a smaller tolerance than obstacles in the lower right.

fallscameron01 (changes requested):
Overall, the structure looks really good, but some subtleties are missing here. The center math and the math that compares the centers needs to be changed. Also there are several things that would give runtime errors, such as calling numpy's size as a function.
Pipeline was updated correctly though.

vision/obstacle/obstacle_tracker.py
fallscameron01: Second sentence is misworded, should read:
You are multiplying the center value by the movement tolerance, but you want to be multiplying the total dimensions of the image by the movement tolerance.

Also, the image dimensions are 1920x1080.

fallscameron01 (approved):
Looks good. Ready to merge.

PR #230, opened 2021-02-26
Feature/frame testing

mat-px:
bench_module.py
-Added accuracy_saveCircleImage function that creates a folder and calls on a function to save image with circles in it
-Created a parser arg "-c" to select if you want to save the images
-Modified so images would only save if they are in frame (also added that to the getcenter)

location.py
-Added saveCircleImage to save the image with circles in the folder created in benchmark
-Added a new blur method
-Added some int() to some variables that were having issues when testing (line 415)


vision/benchmarks/accuracy/bench_module.py
```diff
@@ -164,8 +164,42 @@ def accuracy_get_module_roll(self, region: np.ndarray) -> float:
         """
         return get_module_roll(region)
 
+    def accuracy_saveCircleImage(
```
fallscameron01: Functions that start with accuracy_ will be run by the runall.py, so this should not be in the AccuracyModule class. Also, it looks like this function could just be implemented directly into bench_module_accuracy.

vision/benchmarks/accuracy/bench_module.py
```diff
@@ -164,8 +164,42 @@ def accuracy_get_module_roll(self, region: np.ndarray) -> float:
...
+            The depth image.
+        filename
+            The file name.
+
+        Returns
+        -------
+        None
+        """
+        OUTPUT_FILE = "Circle_Images"
+        try:
```
fallscameron01: Since this function is called every time bench_module_accuracy runs the algorithms on an image, having it throw an error if the directory already exists creates unnecessary overhead. You can instead do something like draw_centers does here at the top of bench_module_accuracy:

`if draw_centers and not os.path.isdir(DRAW_CENTERS_DIR):`
`        os.mkdir(DRAW_CENTERS_DIR)`

vision/benchmarks/accuracy/bench_module.py
```diff
@@ -164,8 +164,42 @@ def accuracy_get_module_roll(self, region: np.ndarray) -> float:
...
+            os.mkdir(OUTPUT_FILE)
+        except FileExistsError:
+            pass
+        self.location.setImg(color_image, depth_image)
+        self.location.getCenter()
```
fallscameron01: Also, since this is called after getCenter has already been called for the image, you don't need to set the image or getCenter() again. You only need the line after this that saves the circle.

vision/benchmarks/accuracy/bench_module.py
```diff
@@ -339,39 +373,33 @@ def bench_module_accuracy(folder: str, quiet_output: bool = False, draw_centers:
                 # calculate execution time
                 exec_time = end_time - start_time
                 f.write(str(exec_time))
-                
+
                 total_time += exec_time
 
+                # Saves image with circles and the module in frame
+                if in_frame and save_circle:
+                    tester.accuracy_saveCircleImage(file, image, depth)
```
fallscameron01: This tester.accuracy... line can just be replaced with the one line from your function above.

`self.location.saveCircleImage(filename, OUTPUT_FILE)`

vision/module/location.py
```diff
@@ -377,6 +385,23 @@ def showCircles(self) -> None:
...
+        None
+        """
+
+        circleImg = np.copy(self.img)
+
+        for x, y, r in self.circles:
+            cv2.circle(circleImg, (x, y), r, (0, 255, 0), 4)
+            cv2.rectangle(circleImg, (x - 5, y - 5), (x + 5, y + 5), (0, 128, 255), -1)
+
+        cv2.imwrite(os.path.join(folder, filename), circleImg)
```
fallscameron01: Instead of using os.path.join here to add the folder to the filename, you can do that in the benchmark and just pass the whole thing as one parameter to this function.

vision/module/location.py
```diff
@@ -2,6 +2,7 @@
 This file contains the ModuleLocation class to find the location of the center of the module in an image.
 """
 
+import os
```
fallscameron01: Don't need to import os. See comment above.

vision/module/location.py
```diff
@@ -377,6 +385,23 @@ def showCircles(self) -> None:
         cv2.waitKey(0)
         cv2.destroyAllWindows()
 
+    def saveCircleImage(self, filename, folder) -> None:
```
fallscameron01: See above comment, can remove the folder parameter. Also, make sure to "type" your parameters. Example:

`def saveCircleImage(self, filename: str) -> None:`

vision/module/location.py
```diff
@@ -377,6 +385,23 @@ def showCircles(self) -> None:
...
+        """
+        Saves image with circles in folder Circle_Images.
+
```
fallscameron01, line 395: Also, need to specify parameters in docstring.

fallscameron01 (changes requested):
Good work for your first pull request. A few changes to be made to clean things up a little. Good work on the circle detection function though.

vision/module/location.py
```diff
@@ -377,6 +384,23 @@ def showCircles(self) -> None:
         cv2.waitKey(0)
         cv2.destroyAllWindows()
 
+    def saveCircleImage(self, file: str) -> None:
+        """
+        Saves image with circles in folder circles.
+
```
fallscameron01, line 390: Still need to specify the parameters in the docstring.

fallscameron01 (changes requested):
Just need to specify the parameters in the circle write function.

fallscameron01 (approved):
Looks good. Good job on your first PR!

PR #231, opened 2021-02-26
Optimize calc method for num_buckets and adjust getCenter and inFrame requirements

mrouie:


vision/benchmarks/accuracy/bench_module.py
```diff
@@ -352,26 +352,33 @@ def bench_module_accuracy(folder: str, quiet_output: bool = False, draw_centers:
                         "CRASH:",crash,
                     )
                 
+
+                marked_image = np.copy(image)
```
fallscameron01: This can be moved inside the an if statement so that it is only copied if needed. Example:

`if draw_centers or (draw_hough_circles and tester.location.circles is not None):`

vision/benchmarks/accuracy/bench_module.py
```diff
@@ -352,26 +352,33 @@ def bench_module_accuracy(folder: str, quiet_output: bool = False, draw_centers:
...
+                        img=marked_image, 
+                        center=(center[0], center[1]), 
+                        radius=1, 
+                        color=(0, 0, 255), 
+                        thickness=2
                     )
 
-                    cv2.imwrite(os.path.join(DRAW_CENTERS_DIR,file), image_copy)
+                # if enabled and available, draw hough circles
+                if draw_hough_circles and tester.location.circles is not None:
```
fallscameron01: Mateusz also added a feature to draw the circles. His uses a function in the location class, so is probably more robust. Make sure you resolve that in the merge conflict properly so that only one version of the draw circles is in the benchmark.

vision/benchmarks/accuracy/bench_module.py
```diff
@@ -352,26 +352,33 @@ def bench_module_accuracy(folder: str, quiet_output: bool = False, draw_centers:
...
+                    for x, y, r in tester.location.circles:
+                        cv2.circle(marked_image, (x, y), r, (0, 255, 0), 4)
+                        cv2.rectangle(marked_image, (x - 5, y - 5), (x + 5, y + 5), (0, 128, 255), -1)
 
+                cv2.imwrite(os.path.join(OUTPUT_IMGS_DIR,file), marked_image)
```
fallscameron01: If you add the if statement I suggested above, this would also need an if statement.

vision/module/location.py
```diff
@@ -253,10 +258,10 @@ def _circleDetection(self) -> np.ndarray:
         gray = cv2.cvtColor(src=self.img, code=cv2.COLOR_RGB2GRAY)
 
         # Guassian Blur
```
fallscameron01: Comment needs to change to reflect updated blur. Documentation here fixed by Mateusz, so just needs to be fixed in the merge conflict.

fallscameron01 (changes requested):
Nice work here. A few issues to address though. The biggest thing is fixing the merge conflict (which you should do before you do anything else). Mateusz had the same feature for drawing the circles. His uses a function he created in location, so I would keep his version. Also a conflict for the updated blur, but that should be essentially the same code. Good work on the updates here, they look really good. Good use of NumPy functions.

fallscameron01 (approved):
Nice job fixing that big merge conflict. Ready to merge to develop. If you wanted to fix that issue we discussed about the ability to save the center without saving the circles, then commit and push that and re-request a review. Otherwise, you can just merge it now.

PR #235, opened 2021-03-05
Filtered NaN return values and added variable search radius

### LINKED ISSUE #225: Ensure get_module_depth() always returns a value.
## Description of Problem

Currently get_module_depth() returns nanfor some images. Should be updated so a float is always returned.

## Requirements

- [x] Find why calculations to get depth sometimes return nan.
- [x] Calculations done to get depth should be updated so nan is not returned.
- [x] Search radius should change dynamically based on distance from center.

Andrew-Gibson42:


vision/module/get_module_depth.py
```diff
@@ -11,7 +11,7 @@
 SEARCH_RADIUS = 20
 
 
-def get_module_depth(depth_image, coordinates):
+def get_module_depth(depth_image, coordinates) -> float:
```
fallscameron01: Make sure to type parameters.

example
`(depth_image: np.ndarray,`

vision/module/get_module_depth.py
```diff
@@ -24,16 +24,44 @@ def get_module_depth(depth_image, coordinates):
...
+    search_radius = SEARCH_RADIUS
+    
+    if direct_center_depth != 0:
+        # 750 chosen as a rough median for depth values
+        search_radius_multipler = direct_center_depth / 750
 
-    depth_values_in_radius = depth_image[y_pos - SEARCH_RADIUS:y_pos + SEARCH_RADIUS, x_pos - SEARCH_RADIUS: x_pos + SEARCH_RADIUS]
+        # inverted because higher depth values means further away
+        if search_radius_multipler > 1:
+            if search_radius_multipler > 2:
```
fallscameron01: This could be changed to an if-elif-else chain rather than nested if statements to make it more clear.
Order would be >2, >1, else

vision/module/get_module_depth.py
```diff
@@ -24,16 +24,44 @@ def get_module_depth(depth_image, coordinates):
...
+        else:
+            search_radius_multipler = 1 + (1 - search_radius_multipler)
+
+        search_radius = round(SEARCH_RADIUS * search_radius_multipler)
+        
+        if search_radius < 10:
+            search_radius = 10
+
+    # must be an integer for array indexing
+    search_radius = int(search_radius)
```
fallscameron01: Can change it to an integer above on the same line where you round.

fallscameron01 (changes requested):
Looks good. Just a little rearranging to do. and a documentation fix.

fallscameron01 (approved):
Looks good.

PR #239, opened 2021-03-17
Feature/better turn

### LINKED ISSUE #172: Equation to make turns based on speed or a better way to turn
## Description of Problem

Currently our turns are pretty much only working at constant speeds, if we fluctuate the speed of the drone, the turns change. Find a better way.

noahklein00:
The turn function now allows for variable input and adjusts the time, degrees, and direction of the turn based off of function variables.


flight/flight.py
```diff
@@ -159,7 +159,7 @@ async def start_flight(comm, drone: System):
                 # Worried about what happens here
             await asyncio.sleep(config.THINK_FOR_S)
             logging.info("Landing the drone")
-            await drone.action.land()
```
ewad3: Understand changing for testing but needs to be changed back for PR

run.py
```diff
@@ -13,13 +13,16 @@
...
+    try:
+        parser: argparse.ArgumentParser = argparse.ArgumentParser()
+        parser.add_argument(
+            "-s", "--simulation", help="using the simulator", action="store_true"
+        )
+        args = parser.parse_args()
+        logging.debug("Simulation flag %s", "enabled" if args.simulation else "disabled")
+        run_threads(args.simulation)
+    except:
+        print("exception thrown all the way back to main")
```
ewad3: Understand changing for testing but needs to be changed back for PR

ewad3 (changes requested):
Other than small changes needed to be made looks good.

flight/utils/movement_controller.py
```diff
@@ -95,25 +96,52 @@ async def move_to(self, drone: System, pylon: LatLon) -> bool:
...
         Turns the drone around the pylon it is currently at
         Parameters:
             Drone(System): Our drone object
         """
+        # counts the number of data points read/sent to the drone per turn
         count: int = 0
+
+        # Gets the velocity of the drone going into the turn
+        async for tel in drone.telemetry.position_velocity_ned():
+            current_vel : double = (tel.velocity.north_m_s**2 + tel.velocity.east_m_s**2)**.5
```
pieperm: This should be annotated as `current_vel: float`

flight/utils/movement_controller.py
```diff
@@ -95,25 +96,52 @@ async def move_to(self, drone: System, pylon: LatLon) -> bool:
...
+            break
+
+        # Constant time
+        yawspeed_d_s = ((-1 * left_turn) * degrees_turned)/time
+
+        forward_m_s = 0
+        right_m_s = 0
+        down_m_s = 0
```
pieperm: Let's annotate these variables as `float` types

flight/utils/movement_controller.py
```diff
@@ -95,25 +96,52 @@ async def move_to(self, drone: System, pylon: LatLon) -> bool:
...
+
         async for tel in drone.telemetry.attitude_euler():
+            # EulerAngle: [roll_deg, pitch_deg, yaw_deg]
+            # Calculates the current angle of the drone
             current: float = (360 + round(tel.yaw_deg)) % 360
+            # Calculate the angle required to turn 180 deg on the first data point
             if count == 0:
-                temp = (current + 180) % 360
+                temp = (current + ((-1 * left_turn) * degrees_turned)) % 360
+                midpoint = (current + ((-1 * left_turn) * degrees_turned / 2)) % 360
```
pieperm: More `float` types to annotate

flight/utils/movement_controller.py
```diff
@@ -95,25 +96,52 @@ async def move_to(self, drone: System, pylon: LatLon) -> bool:
...
+            # Sends signal to the drone to turn. **No need to send this multiple times
+            #   unless the value is changing each data point
             await drone.offboard.set_velocity_body(
-                sdk.offboard.VelocityBodyYawspeed(5, -3, -0.1, -60)
+                sdk.offboard.VelocityBodyYawspeed(forward_m_s, right_m_s, down_m_s, yawspeed_d_s)
             )
             # await asyncio.sleep(config.FAST_THINK_S)
+            # Finds the difference between the current angle and desired angle
             val = abs(current - temp)
+            mid_val = abs(current - midpoint)
```
pieperm: More `float` types to annotate

pieperm (changes requested):
Lots of comments that do a great job explaining what's happening. Could use changes with type annotations on a few variables.

flight/utils/movement_controller.py
```diff
@@ -95,25 +96,52 @@ async def move_to(self, drone: System, pylon: LatLon) -> bool:
                 return True
             count += 1
 
-    async def turn(self, drone: System) -> bool:
+    async def turn(self, drone: System, degrees_turned = 180: int, left_turn = True: bool, time = 3.5: float) -> bool:
```
pieperm: The type annotations need to be changed here such that the type comes before the default parameter:
```
async def turn(self, drone: System, degrees_turned: int = 180, left_turn: bool = True, time: float = 3.5) -> bool:
```

pieperm: Another possible issue I noticed is that adding these arguments is not reflected when the function is called. That is, in `states/early_laps.py`, the call to `mover.turn()` is only passed a `drone`. Are we planning on using the default arguments for the pylon laps?

pieperm (changes requested):
A couple more things

flight/utils/movement_controller.py
pieperm: Additionally, the new parameters should be described in the docstring (along with the existing `drone` description)

PR #240, opened 2021-03-17
Test module roll

### LINKED ISSUE #214: Add unit test for module roll
## Description of Problem
Add a unit test to test the get_module_roll() function.

## Requirements
- [x] Should be separate from the module orientation unit test due these being different functions
- [x] Test to ensure function works when passed correct data type.
- [x] Test to ensure correct output returned.

mat-px:


vision/benchmarks/accuracy/bench_module.py
```diff
@@ -318,7 +318,7 @@ def bench_module_accuracy(
                     try:
                         bounds = tester.accuracy_getModuleBounds(
                             (1920, 1080), center, depth_val
-                        )
+                        )  # Switched parameters
```
fallscameron01: Shouldn't leave your changes as comments here since this clutters up the code.

vision/benchmarks/accuracy/bench_module.py
```diff
@@ -334,8 +334,8 @@ def bench_module_accuracy(
                 if depth_val != 0 and not crash:
                     try:
                         bound_region = image[
-                            bounds[0][1] : bounds[3][1], bounds[0][0] : bounds[3][0], :
-                        ]
+                            bounds[0][1] : bounds[3][1], bounds[0][0] : bounds[2][0], :
+                        ]  # Fixed bounds?
```
fallscameron01: Same thing here.

vision/module/module_bounding.py
```diff
@@ -34,7 +34,8 @@ def getModuleBounds(dimensions, center, depth):
     x, y = center
     vert_res, horiz_res = dimensions
 
-    vert_angle = np.degrees(np.arctan((MODULE_HEIGHT / 2) / depth))
+    # switched height and width so it worked while testing
```
fallscameron01: And here.

vision/module/module_bounding.py
```diff
@@ -34,7 +34,8 @@ def getModuleBounds(dimensions, center, depth):
...
+    vert_angle = np.degrees(np.arctan((MODULE_WIDTH / 2) / depth))
```
fallscameron01: For this, I think the width and height should be flipped for the vert angle and horizontal angle since the module is taller than it is wide.

vision/pipeline.py
```diff
@@ -123,8 +123,8 @@ def run(self, prev_state):
                 )  # overestimate of bounds
                 roll = get_module_roll(
                     color_image[
-                        bounds[0][1] : bounds[3][1], bounds[0][0] : bounds[3][0], :
-                    ]
+                        bounds[0][1] : bounds[3][1], bounds[0][0] : bounds[2][0], :
+                    ]  # Fixed bounds?
```
fallscameron01: Another change comment.

vision/unit_tests/test_module.py
```diff
@@ -15,7 +15,9 @@
 from vision.module.in_frame import ModuleInFrame as mif
 from vision.module.location import ModuleLocation
 from vision.module.region_of_interest import region_of_interest
-from vision.module.module_orientation import get_module_orientation
+from vision.module.module_orientation import *
+from vision.module.module_bounding import getModuleBounds
+from vision.module.get_module_depth import *
```
fallscameron01: Should specify the specific modules you are importing. Using stars is unclear.

Unrelated:
Also, there was a big refactor that got pulled into develop, so getModuleBounds is now get_module_bounds. Will need to change that throughout your benchmark here to reflect that. Same thing with getCenter being changed to get_center, which needs to be changed throughout this file.
The benchmark was also refactored, which is what is causing a merge conflict.

vision/unit_tests/test_module.py
```diff
@@ -363,10 +363,96 @@ def test_return(self):
...
+            "2020-02-29_15.36.15.167565": 13.5,
+            "2020-02-29_15.40.46.652448": 14,
+            "2020-02-29_15.40.48.862847": 13,
+            "2020-02-29_15.40.40.444538": 9,
+            "2020-02-29_15.40.54.564617": 10,
+        }
+
+        for current_file in estimates.keys():
+            IMAGE_SIZE = [1920, 1080]
+            # grabs and loads image file (assumes all files are in order of color1, depth1, color2, depth2, etc.)
```
fallscameron01: > assumes all files are in order of color1, depth1...

I'm not sure what this is meant to imply. The files don't need to be in order and since they are handpicked images they will have the format you specified.

vision/unit_tests/test_module.py
```diff
@@ -363,10 +363,96 @@ def test_return(self):
...
+            current_image_file = os.path.join(img_dir, current_file) + "-colorImage.jpg"
+            current_depth_file = os.path.join(img_dir, current_file) + "-depthImage.npy"
+            # loads images
+            current_image_file = cv2.imread(current_image_file)
+            current_depth_file = np.load(current_depth_file)
+
+            # sets images to get center, depth_val, and bounds
+            # works given the center value is correct from ModuleLocation()
+            loc.setImg(current_image_file, current_depth_file)
+            center = loc.getCenter()
```
fallscameron01: Since this is a test for get_module_roll, you shouldn't be relying on too many other functions succeeding. Should be using an estimated value for the center and pull the depth directly from the depth image via indexing that center rather than using get_center and get_module_depth. get_module_bounds should stay since the roll is heavily dependent on it and there is no good way to estimate it.

vision/unit_tests/test_module.py
```diff
@@ -363,10 +363,96 @@ def test_return(self):
...
+        }
+
+        for current_file in estimates.keys():
+            IMAGE_SIZE = [1920, 1080]
+            # grabs and loads image file (assumes all files are in order of color1, depth1, color2, depth2, etc.)
+            current_image_file = os.path.join(img_dir, current_file) + "-colorImage.jpg"
+            current_depth_file = os.path.join(img_dir, current_file) + "-depthImage.npy"
+            # loads images
+            current_image_file = cv2.imread(current_image_file)
+            current_depth_file = np.load(current_depth_file)
```
fallscameron01: Should use different variable names instead of reassigning an ndarray variable to a string variable.

vision/unit_tests/test_module.py
```diff
@@ -363,10 +363,96 @@ def test_return(self):
...
+        ----------
+        enclosing_region: ndarray
+            The depth image.
+        module_roll: np.float64
+            Calculated float angle.
+        """
+
+        IMAGE_SIZE = [1080, 1920, 3]
+
+        enclosing_region = np.zeros(IMAGE_SIZE, dtype=np.uint8)
```
fallscameron01, line 390: Good catch adding the color channel to IMAGE_SIZE. Looks like it doesn't do this for the params or return functions for TestModuleOrientation, if you could also fix that.

vision/unit_tests/test_module.py
```diff
@@ -363,10 +363,96 @@ def test_return(self):
...
+        self.assertIs(type(module_roll), np.float64)
+
+    def test_module_roll(self):
+        img_dir = os.path.join(gparent_dir, "vision_images/module/Feb29")
+        files = os.listdir(img_dir)
+        loc = ModuleLocation()
+
+        estimates = {
+            # FORMAT: "name_of_file" : estimated float // COULD BE WILDLY OFF, just guesses as of right now
+            "2020-02-29_15.36.15.167565": 13.5,
```
fallscameron01: Can put the estimated center in list with roll for the value. See test_region_of_interest for example.

fallscameron01 (changes requested):
You are getting a merge conflict due to the benchmark refactor. MAKE SURE TO RESOLVE THIS FIRST (or it will make life difficult). Looks like the actual merge conflict is in the module benchmark, which was heavily reworked. Make sure you only bring in the code you changed here and leave the rest the same as what is on develop.
The refactor also changed the names of getCenter to get_center and getModuleBounds to get_module_bounds, so you should change that where needed.

For the actual unit test, you did a good job of implementing it. Main thing is that you are calling on get_center and get_module_depth, which adds extra points of failure to the test. Instead, you should use estimated values for the center and use that estimate to index the depth image for the depth value. Since the unit test is meant to test the roll function, it is important to remove the center and depth functions so it can't fail there and then throw off the roll values. The call to get_module_bounds is fine though since there is not a good way to estimate it.
Only other thing is small changes throughout.

vision/benchmarks/accuracy/bench_module.py
```diff
@@ -39,14 +39,12 @@ def accuracy_is_in_frame(
     ) -> bool:
         """
         Measuring accuracy of is_in_frame() from ModuleLocation class.
-
```
fallscameron01: Looks like when you were fixing the merge conflict, you removed a bunch of new lines in docstrings. Should go back and put these back in because it messes up the readability.

fallscameron01: Looks like this happened in every docstring in bench_module.py.

fallscameron01: Also it looks like you didn't carry your change to the module bounds indexing over when you resolved the merge conflict.

vision/unit_tests/test_module.py
```diff
@@ -363,10 +361,87 @@ def test_return(self):
...
+            current_depth_file = os.path.join(img_dir, current_file) + "-depthImage.npy"
+            # Loads images
+            colorImage = cv2.imread(current_image_file)
+            depthImage = np.load(current_depth_file)
+
+            center = estimates[current_file][1]
+
+            # Sets a rough boundary around the module (angles are similar using get_module_bounds)
+            bound = [center[1] - 250, center[1] + 250, center[0] - 180, center[0] + 180]
+            colorImage = colorImage[bound[0] : bound[1], bound[2] : bound[3], :]
```
fallscameron01: Should use another variable name here. Since you are taking only a portion of the color image, it is strange to still call it the color image. Should be separate variable to increase readability.

vision/unit_tests/test_module.py
```diff
@@ -363,10 +361,87 @@ def test_return(self):
...
+
+            estimated_degrees = get_module_roll(colorImage)
+            calculated_degrees = estimates[current_file][0]
```
fallscameron01: Variable names are backwards here. You are assigning estimated_degrees to the value calculated by the function and you are assigning calculated_degrees to your estimate.

vision/unit_tests/test_module.py
```diff
@@ -363,10 +361,87 @@ def test_return(self):
...
+
+        self.assertIs(type(enclosing_region), np.ndarray)
+        self.assertIs(type(module_roll), np.float64)
+
+    def test_module_roll(self):
+        img_dir = os.path.join(gparent_dir, "vision_images/module/Feb29")
+        files = os.listdir(img_dir)
+
+        estimates = {
+            # FORMAT: "name_of_file" : [estimated float, (estimated center)]
```
fallscameron01: > estimated float

Should specify that this is the roll value.

vision/unit_tests/test_module.py
```diff
@@ -363,10 +361,87 @@ def test_return(self):
...
+            # Sets a rough boundary around the module (angles are similar using get_module_bounds)
+            bound = [center[1] - 250, center[1] + 250, center[0] - 180, center[0] + 180]
+            colorImage = colorImage[bound[0] : bound[1], bound[2] : bound[3], :]
+
+            estimated_degrees = get_module_roll(colorImage)
+            calculated_degrees = estimates[current_file][0]
+
+            self.assertTrue(
+                estimated_degrees * 0.8 <= calculated_degrees <= estimated_degrees * 1.2
+            )  # within ±20%
```
fallscameron01: 20% is quite a large interval. 20% on either side would mean a 40% range, which is quite large. Should lower this from 20 to 10.

vision/unit_tests/test_module.py
```diff
@@ -363,10 +361,87 @@ def test_return(self):
...
+            "2020-02-29_15.40.54.564617": [8, (435, 650)],
+        }
+
+        for current_file in estimates.keys():
+            # Selects image
+            current_image_file = os.path.join(img_dir, current_file) + "-colorImage.jpg"
+            current_depth_file = os.path.join(img_dir, current_file) + "-depthImage.npy"
+            # Loads images
+            colorImage = cv2.imread(current_image_file)
+            depthImage = np.load(current_depth_file)
```
fallscameron01: One other small thing is that we are trying to move towards using snake case (instead of camel case) for everything except class names, so these should be changed.

Should be:
`color_image`
`depth_image`

fallscameron01: Looks like this is the only place you used camel case, so these are the only ones that need to be changed. (unless I missed some somewhere).

vision/unit_tests/test_module.py
fallscameron01: Calculation below is in correct order for the variable names, so you just need to change the assignments.

fallscameron01 (changes requested):
Looks like you didn't resolve the merge conflict properly. You ended up removing a lot of blank lines in docstrings in bench_module.py which hurt the readability. You also did not bring your changes over in bench_module.py that were on the test_module_roll branch. Make sure you resolve merge conflicts in an editor instead of copy-pasting, because that can cause issues and is harder to keep track of.

Other than that, just some changes to variable names, a change to your interval for considering the calculated result a success, and a comment change.

vision/unit_tests/test_module.py
```diff
@@ -363,10 +361,87 @@ def test_return(self):
...
+            "2020-02-29_15.40.54.564617": [8, (435, 650)],
+        }
+
+        for current_file in estimates.keys():
+            # Selects image
+            color_image = os.path.join(img_dir, current_file) + "-colorImage.jpg"
+            depth_image = os.path.join(img_dir, current_file) + "-depthImage.npy"
+            # Loads images
+            colorImage = cv2.imread(color_image)
+            depthImage = np.load(depth_image)
```
fallscameron01: Above issue not resolved. I think you changed the name of the wrong variables. Talking about these lines.
Should be:

> color_file = os.path.join(img_dir, current_file) + "-colorImage.jpg"
            depth_file = os.path.join(img_dir, current_file) + "-depthImage.npy"
            # Loads images
            color_image = cv2.imread(color_file)
            depth_image = np.load(depth_file)

fallscameron01: Original comment:
One other small thing is that we are trying to move towards using snake case (instead of camel case) for everything except class names, so these should be changed.

fallscameron01 (changes requested):
Unresolved issue from previous review needs to be addressed.

mat-px:
My bad, didn't know what snake case and camel case. I also removed the depth images stuff since it isn't used.

fallscameron01 (approved):
Nice work! Thanks for fixing those issues.

PR #241, opened 2021-03-17
Feature/exit state

cjhr95:
Addition of try-except block to handle Ctrl-C exit of flight and any errors that occur. Also added logging statements to keep track of errors.


flight/utils/movement_controller.py
```diff
@@ -203,11 +203,9 @@ async def move_to_takeoff(self, drone: System, takeoff_location: LatLon) -> None
             # if the x and y values are close enough (2m) to the original position * precision
             # if inside the circle, move on to the next
             # if outside of the circle, keep running to you get inside
-            if (
-                abs(x) <= reference_x * config.POINT_PERCENT_ACCURACY
-                and abs(y) <= reference_y * config.POINT_PERCENT_ACCURACY
-            ):
+            if abs(x) <= reference_x * 0.15 and abs(y) <= reference_y * 0.15:
```
ewad3: I understand what you are doing here, but on the physical drone having to high of point accuracy causes the drone to "miss" its target and fly back and forth like it did at the flight test while trying to land.

ewad3 (dismissed):
Everything looks good. We just will need to keep an eye on the landing.

run.py
```diff
@@ -61,14 +61,24 @@ def run_threads(sim: bool) -> None:
                 logging.error("Flight process terminated, restarting")
                 flight_process: Process = init_flight(flight_args)
                 flight_process.start()
-    except KeyboardInterrupt:
+    except KeyboardInterrupt or AttributeError:
```
pieperm: I'm not sure if this is the right syntax for catching multiple errors. See https://stackoverflow.com/questions/6470428/catch-multiple-exceptions-in-one-line-except-block

Let me know if I'm wrong about this.

run.py
```diff
@@ -61,14 +61,24 @@ def run_threads(sim: bool) -> None:
...
+            logging.info("Non-fatal AttributeError occurs, land still possible")
+            flight_process: Process = init_flight(flight_args)
+            flight_process.start()
+        except EOFError:
+            logging.info("Additional EOF error detected - lack of inputs on restart")
+            comm_obj.set_state("hold")
+            comm_obj.set_state("land")
+            flight_process: Process = init_flight(flight_args)
+            flight_process.start()
+        except TimeoutError or flight.DroneNotFoundError:
```
pieperm: Same issue as mentioned before

pieperm (changes requested):
I'm a bit hesitant to make changes to `run.py` before a flight test since it is so crucial. I think it might be a good idea to keep this on its own branch up through the flight test (similarly to how the unit test refactor will be kept on its own branch). We can run a test with this by switching to the branch during the flight test, and then if everything runs smoothly, we can merge it afterwards. What do you think about this approach?

cjhr95:
I can wait to merge the pull request, the points Michael made are valid. I'm not too concerned with merging the request, I'd just like to make sure everything looks decent

pieperm (dismissed):
Looks ready to test at the flight test! If all goes well, we can merge it after

pieperm (approved):

PR #243, opened 2021-03-20
Refactor for Integration Testing

pieperm:
A few features here:
- New `FlightManager` class for flight startup procedures that were originally in `run.py`
- The `StateSettings` class replaces some of the global settings in `config.py`. The config was is meant to store constant values, and the state settings would change based on what we were testing for (at flight tests)
- Each state class (`EarlyLaps`, `Takeoff`, etc.) now has access to a `StateSettings` object via `self.state_settings` (which is originally set up in the base `State` class). This is generally set up in `run.py` and other segment tests, and then simply passed through the state machine.
- There are also a few example segment tests in the `segment_tests` directory
- Additionally there is a new `ModuleDetection` state for testing purposes


pieperm:
Updates since the initial comment:
- The term "segment tests" has been renamed to a more accurate term: integration tests. These are kept in the `integration_tests` directory
- New settings for return laps, title, description, and more
- Fixed import issues

ewad3 (approved):
Looks really good and ready for testing.

PR #247, opened 2021-04-09
Feature/pipelinetest

### LINKED ISSUE #221: Ensure module algorithms work together
## Description of Problem

Ensure that the module detection algorithms work properly together, and fix any issues in running module detection algorithms from the pipeline.

## Requirements

- [x] Ensure pipeline can run module detection algorithms together without errors
- [x] Ensure datatypes are correct between module classes and functions
- [x] Fix any runtime errors when executing module detection algorithms from pipeline

### LINKED ISSUE #238: Handle crashes gracefully in the pipeline.
## Description of Problem

Currently many some function throw errors under certain circumstances. Crashing should be handled in the pipeline to ensure pipeline can recover and continue processing image if possible.

## Requirements

- [x] Handle crashes in the pipeline using try-except blocks.
- [x] Allow algorithms to continue as much as possible. (example: if module orientation crashes, we can still try to get the module roll)
- [x] Ensure pipeline does not attempt to continue if further processing would be unreasonable. (example: don't try to get the module depth if the module center crashed)
- [x] Add new bounding box type "crash" to differentiate between no bounding boxes detected vs algorithm crash.

### LINKED ISSUE #245: Pipeline failure flags
## Description of Problem

In order to differentiate between crash and no bounding boxes being detected, failure flags should be implemented in the pipeline. These flags will signify that an algorithm failed to run to completion.

## Requirements

- [x] Add failure flags. These can be implemented as a new separate object type.
- [x] Flags should be algorithm-specific. For example, each module algorithm (is_in_frame(), get_center(), etc.) should have its own flag.
- [x] Flag object should be specific to pipeline state. For example, would not show obstacle detection failure flag during module detection state.
- [x] Flags should be saved per run. Need a queue or similar to store flags every time run is called.
- [x] Flags should be able to be outputted as string so they can be logged.

Andrew-Gibson42:
Fixed some errors with bag files in various algorithms used by the pipeline, added failure flags objects and logged algorithm outcome in a list of failure flags objects as a member variable of the pipeline class

### DIFF
--- a/vision/failure_flags.py
+++ b/vision/failure_flags.py
@@ -9,6 +9,9 @@ class FailureFlags:
     """
     def __init__(self):
         pass
+
+    def __str__(self):
+        return "INVALID STATE"
     
 class ObstacleDetectionFlags(FailureFlags):
     """
--- a/vision/failure_flags.py
+++ b/vision/failure_flags.py
@@ -1,0 +1,58 @@
+
+class FailureFlags:
+    """
+    Abstract class used by the pipeline to track which algorithm failed
+    in the event of the program crashing
+
+    For each attribute, a value of True signals that the algorithm
+    ran to completion without any problems
+    """
+    
+class ObstacleDetectionFlags(FailureFlags):
+    """
+    Class used by the pipeline to track which algorithm failed
+    in the event of the program crashing during obstacle detection ("early_laps")
+    """
+    def __init__(self):
+        self.obstacle_finder = True
+        self.obstacle_tracker = True
+
+    def __str__(self):
+        return ("obstacle_finder: " + ("OK" if self.obstacle_finder else "ERROR") +
+               ", obstacle_tracker: " + ("OK" if self.obstacle_tracker else "ERROR"))
+
+class TextDetectionFlags(FailureFlags):
+    """
+    Class used by the pipeline to track which algorithm failed
+    in the event of the program crashing during text detection
+    """
+    def __init__(self):
+        self.detect_russian_word = True
+
+    def __str__(self):
+        return "detect_russian_word: " + ("OK" if self.detect_russian_word else "ERROR")
+
+class ModuleDetectionFlags(FailureFlags):
+    """
+    Class used by the pipeline to track which algorithm failed
+    in the event of the program crashing during module detection
+    """
+    def __init__(self):
+        self.setImg = True
+        self.isInFrame = True
+        self.getCenter = True
+        self.get_module_depth = True
+        self.region_of_interest = True
+        self.get_module_orientation = True
+        self.getModuleBounds = True
+        self.get_module_roll = True
+
+    def __str__(self):
+        return ("setImg: " + ("OK" if self.setImg else "ERROR") +
+               ", isInFrame: " + ("OK" if self.isInFrame else "ERROR") +
+               ", getCenter: " + ("OK" if self.getCenter else "ERROR") +
+               ", get_module_depth: " + ("OK" if self.get_module_depth else "ERROR") +
+               ", region_of_interest: " + ("OK" if self.region_of_interest else "ERROR") +
+               ", get_module_orientation: " + ("OK" if self.get_module_orientation else "ERROR") +
+               ", getModuleBounds: " + ("OK" if self.getModuleBounds else "ERROR") +
+               ", get_module_roll: " + ("OK" if self.get_module_roll else "ERROR"))
--- a/vision/failure_flags.py
+++ b/vision/failure_flags.py
@@ -7,6 +7,8 @@ class FailureFlags:
     For each attribute, a value of True signals that the algorithm
     ran to completion without any problems
     """
+    def __init__(self):
+        pass
     
 class ObstacleDetectionFlags(FailureFlags):
     """
@@ -38,21 +40,21 @@ class ModuleDetectionFlags(FailureFlags):
     in the event of the program crashing during module detection
     """
     def __init__(self):
-        self.setImg = True
-        self.isInFrame = True
-        self.getCenter = True
+        self.set_img = True
+        self.is_in_frame = True
+        self.get_center = True
         self.get_module_depth = True
         self.region_of_interest = True
         self.get_module_orientation = True
-        self.getModuleBounds = True
+        self.get_module_bounds = True
         self.get_module_roll = True
 
     def __str__(self):
-        return ("setImg: " + ("OK" if self.setImg else "ERROR") +
-               ", isInFrame: " + ("OK" if self.isInFrame else "ERROR") +
-               ", getCenter: " + ("OK" if self.getCenter else "ERROR") +
+        return ("setImg: " + ("OK" if self.set_img else "ERROR") +
+               ", is_in_frame: " + ("OK" if self.is_in_frame else "ERROR") +
+               ", get_center: " + ("OK" if self.get_center else "ERROR") +
                ", get_module_depth: " + ("OK" if self.get_module_depth else "ERROR") +
                ", region_of_interest: " + ("OK" if self.region_of_interest else "ERROR") +
                ", get_module_orientation: " + ("OK" if self.get_module_orientation else "ERROR") +
-               ", getModuleBounds: " + ("OK" if self.getModuleBounds else "ERROR") +
+               ", get_module_bounds: " + ("OK" if self.get_module_bounds else "ERROR") +
                ", get_module_roll: " + ("OK" if self.get_module_roll else "ERROR"))
--- a/vision/module/module_bounding.py
+++ b/vision/module/module_bounding.py
@@ -32,7 +32,7 @@ def getModuleBounds(dimensions, center, depth):
     list - list of the four tuple vertices.
     """
     x, y = center
-    vert_res, horiz_res, channels = np.shape(dimensions)
+    vert_res, horiz_res = dimensions
 
     vert_angle = np.degrees(np.arctan((MODULE_HEIGHT / 2) / depth))
     horiz_angle = np.degrees(np.arctan((MODULE_HEIGHT / 2) / depth))
--- a/vision/module/module_bounding.py
+++ b/vision/module/module_bounding.py
@@ -32,7 +32,7 @@ def getModuleBounds(dimensions, center, depth):
     list - list of the four tuple vertices.
     """
     x, y = center
-    vert_res, horiz_res = dimensions
+    vert_res, horiz_res, channels = np.shape(dimensions)
 
     vert_angle = np.degrees(np.arctan((MODULE_HEIGHT / 2) / depth))
     horiz_angle = np.degrees(np.arctan((MODULE_HEIGHT / 2) / depth))
--- a/vision/module/module_orientation.py
+++ b/vision/module/module_orientation.py
@@ -81,16 +81,11 @@ def get_module_roll(enclosing_region: np.ndarray) -> float:
 
     np_rectangles = np.asarray(rectangles)
 
+    # NOTE: this try block exists such that the [:, 2] slice of
+    #         np_rectangles can be attempted without risking
+    #         crashing the algorithm and logging the failure flag
     try:
         angles = np_rectangles[:, 2]
-
-        # if the angle is absurd, it won't be taken
-        #   if the module is truly rolled at over 45 degrees, minAreaRect
-        #   should simply choose a different rectangle or axis
-        # mask = np.where(abs(angles) < 45)
-        # NOTE: the above claim was not the case. Angles with an absolute value
-        #       above 45 are common, and should probably be used to not get a 0
-        #       value in most cases
         
         roll = np.mean(angles)
     except:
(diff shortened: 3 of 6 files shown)


vision/failure_flags.py
```diff
@@ -0,0 +1,58 @@
+
+class FailureFlags:
+    """
+    Abstract class used by the pipeline to track which algorithm failed
+    in the event of the program crashing
+
+    For each attribute, a value of True signals that the algorithm
+    ran to completion without any problems
+    """
```
fallscameron01, line 9: Should add an init function so this isn't just an empty class. Can be as simple as having a pass statement.

vision/failure_flags.py
```diff
@@ -0,0 +1,58 @@
...
+    Class used by the pipeline to track which algorithm failed
+    in the event of the program crashing during obstacle detection ("early_laps")
+    """
+    def __init__(self):
+        self.obstacle_finder = True
+        self.obstacle_tracker = True
+
+    def __str__(self):
+        return ("obstacle_finder: " + ("OK" if self.obstacle_finder else "ERROR") +
+               ", obstacle_tracker: " + ("OK" if self.obstacle_tracker else "ERROR"))
```
fallscameron01, line 27: Nice job on these string functions.

vision/module/module_bounding.py
```diff
@@ -32,7 +32,7 @@ def get_module_bounds(dimensions, center, depth):
     list - list of the four tuple vertices.
     """
     x, y = center
-    vert_res, horiz_res = dimensions
+    vert_res, horiz_res, channels = np.shape(dimensions)
```
fallscameron01: Since the channels isn't being used, you can do this instead:
`vert_res, horiz_res, _ = np.shape(dimensions)`

just so you aren't creating a variable and then not using it.

vision/module/module_orientation.py
```diff
@@ -80,13 +80,21 @@ def get_module_roll(enclosing_region: np.ndarray) -> float:
...
 
     np_rectangles = np.asarray(rectangles)
-    angles = np_rectangles[:, 2]
 
-    # if the angle is absurd, it won't be taken
-    #   if the module is truly rolled at over 45 degrees, minAreaRect
-    #   should simply choose a different rectangle or axis
-    mask = np.where(abs(angles) < 45)
-    roll = np.mean(angles[mask])
+    try:
```
fallscameron01, line 87: Since we now have failure flags for the pipeline, this try block can be removed form within the algorithm. Unless I'm misunderstanding this and a crash would actually mean 0 roll. But otherwise you can just have 0 be a default value in the try-except block in the pipeline. (along with the flag)

vision/pipeline.py
```diff
@@ -88,44 +93,92 @@ def run(self, prev_state):
...
         except Empty:
             state = prev_state
-
         ##
+
         bboxes = []
 
+        flags = {}
+
+        state = "module_detection"
```
fallscameron01: Should remove this state assignment here since it would reassign it every time to module detection

vision/pipeline.py
```diff
@@ -88,44 +93,92 @@ def run(self, prev_state):
             state = self.flight_communication.get_nowait()
         except Empty:
             state = prev_state
-
         ##
+
         bboxes = []
 
+        flags = {}
```
fallscameron01: Can assign flags to the base flag class here to make this more clear.

vision/pipeline.py
```diff
@@ -88,44 +93,92 @@ def run(self, prev_state):
...
+
+            try:
+                bboxes = self.obstacle_finder.find(color_image, depth_image)
+                self.obstacle_tracker.update(bboxes)
+            except:
+                flags.obstacle_finder = False
 
-            obstacle_tracker.update(bboxes)
-            bboxes = obstacle_tracker.getPersistentObstacles()
+            try:
```
fallscameron01: For the cases where one algorithm is depends on another, you should check whether the previous algorithm crashed before trying to run the dependent algorithm.

vision/pipeline.py
```diff
@@ -88,44 +93,92 @@ def run(self, prev_state):
...
-                        bounds[0][1] : bounds[3][1], bounds[0][0] : bounds[2][0], :
-                    ]
-                )  # roll of module
+            if inFrame:
+                try:
+                    center = self.module_location.getCenter()  # center of module in image
+                except:
+                    flags.getCenter = False
+
+                try:
```
fallscameron01, line 135: Same thing with a lot of the module algorithms. These should flow mostly linearly except for where get_module_bounds doesn't depend on region/orientation.

vision/pipeline.py
```diff
@@ -88,44 +93,92 @@ def run(self, prev_state):
...
+                
+                try:
+                    orientation = get_module_orientation(
+                        region
+                    )  # x and y tilt of module
+                except:
+                    flags.get_module_orientation = False
+
+                try:
+                    bounds = getModuleBounds(
```
fallscameron01: getModuleBounds was changed to get_module_bounds

vision/failure_flags.py
```diff
@@ -0,0 +1,58 @@
...
+        return "detect_russian_word: " + ("OK" if self.detect_russian_word else "ERROR")
+
+class ModuleDetectionFlags(FailureFlags):
+    """
+    Class used by the pipeline to track which algorithm failed
+    in the event of the program crashing during module detection
+    """
+    def __init__(self):
+        self.setImg = True
+        self.isInFrame = True
```
fallscameron01: Speaking of, we are trying to move towards a coding standard of using camel case for classes, but snake case for everything else, so these variables should be changed to snake case.

vision/failure_flags.py
```diff
@@ -0,0 +1,58 @@
...
+        self.get_module_roll = True
+
+    def __str__(self):
+        return ("setImg: " + ("OK" if self.setImg else "ERROR") +
+               ", isInFrame: " + ("OK" if self.isInFrame else "ERROR") +
+               ", getCenter: " + ("OK" if self.getCenter else "ERROR") +
+               ", get_module_depth: " + ("OK" if self.get_module_depth else "ERROR") +
+               ", region_of_interest: " + ("OK" if self.region_of_interest else "ERROR") +
+               ", get_module_orientation: " + ("OK" if self.get_module_orientation else "ERROR") +
+               ", getModuleBounds: " + ("OK" if self.getModuleBounds else "ERROR") +
```
fallscameron01: Looks like you are referring to this with the old function name throughout your code, so make sure to got through and update that.

vision/pipeline.py
```diff
@@ -88,44 +93,92 @@ def run(self, prev_state):
...
+                        color_image, center, depth
+                    )  # overestimate of bounds
+                except:
+                    flags.getModuleBounds = False
+
+                try:
+                    roll = get_module_roll(
```
fallscameron01: Need to make sure all of the results of these algorithms have default values and aren't local variables so that they can always be used to construct a bounding box.

vision/pipeline.py
```diff
@@ -88,44 +93,92 @@ def run(self, prev_state):
...
+                        color_image[
+                            bounds[0][1] : bounds[3][1], bounds[0][0] : bounds[2][0], :
+                        ]
+                    )  # roll of module
+                except:
+                    flags.get_module_roll = False
 
                 # construct boundingbox for the module
                 box = BoundingBox(bounds, ObjectType.MODULE)
```
fallscameron01: I would add here that we wouldn't want to construct a module bounding box if the get_center function crashed, so you should add an if statement to check the flag.

vision/failure_flags.py
```diff
@@ -0,0 +1,58 @@
...
+        self.get_module_depth = True
+        self.region_of_interest = True
+        self.get_module_orientation = True
+        self.getModuleBounds = True
+        self.get_module_roll = True
+
+    def __str__(self):
+        return ("setImg: " + ("OK" if self.setImg else "ERROR") +
+               ", isInFrame: " + ("OK" if self.isInFrame else "ERROR") +
+               ", getCenter: " + ("OK" if self.getCenter else "ERROR") +
```
fallscameron01: getCenter and isInFrame also changed to snake case.

vision/pipeline.py
```diff
@@ -88,44 +93,92 @@ def run(self, prev_state):
...
+                try:
+                    orientation = get_module_orientation(
+                        region
+                    )  # x and y tilt of module
+                except:
+                    flags.get_module_orientation = False
+
+                try:
+                    bounds = getModuleBounds(
+                        color_image, center, depth
```
fallscameron01: You can probably just pass the result of np.shape(color_image)[:2] to the get_module_bounds instead of passing the whole color image which isn't being used.

vision/pipeline.py
```diff
@@ -88,44 +93,92 @@ def run(self, prev_state):
...
         elif state == "module_detection":  # locating module
-            self.module_location.set_img(color_image, depth_image)
+            flags = ModuleDetectionFlags()
+            try:
+                self.module_location.setImg(color_image, depth_image)
+            except:
+                flags.setImg = False
+
+            try:
+                inFrame = self.module_location.isInFrame()
```
fallscameron01: Should change to snake case.

fallscameron01 (changes requested):
Looks good overall. Lots of little changes and formatting changes. Also, for algorithms that are dependent on another algorithm succeeding, you should be checking the failure flag to ensure it ran to completion before attempting to run the dependent algorithm. Example: shouldn't run obstacle tracker if obstacle finder crashed. Needs to be done for the module algorithms too.

vision/pipeline.py
Andrew-Gibson42: For testing purposes, forgot to take it out-- good catch

vision/module/module_orientation.py
Andrew-Gibson42: The try block is in place so that we can attempt to slice np_rectangles with [:, 2] without a crash if the image is the incorrect shape-- I could set roll to 0 in the pipeline instead, but then it would also set the failure flag, and I'm not sure if that's desirable in this case.

vision/pipeline.py
```diff
@@ -88,54 +95,116 @@ def run(self, prev_state):
...
+
         if state == "early_laps":  # navigation around the pylons
-            bboxes = self.obstacle_finder.find(color_image, depth_image)
+            flags = ObstacleDetectionFlags()
 
-            obstacle_tracker.update(bboxes)
-            bboxes = obstacle_tracker.getPersistentObstacles()
+            try:
+                bboxes = self.obstacle_finder.find(color_image, depth_image)
+                self.obstacle_tracker.update(bboxes)
```
fallscameron01: Should move the update function to be in the obstacle tracker try block.

vision/pipeline.py
```diff
@@ -88,54 +95,116 @@ def run(self, prev_state):
...
-                bboxes.append(box)
+            flags = ModuleDetectionFlags()
+            try:
+                self.module_location.setImg(color_image, depth_image)
+            except:
+                flags.set_img = False
+
+            if flags.set_img:
+                try:
+                    inFrame = self.module_location.is_in_frame()
```
fallscameron01: Should change this variable to snake case.

vision/pipeline.py
```diff
@@ -88,54 +95,116 @@ def run(self, prev_state):
...
+                    orientation = (0, 0)
+                    bounds = np.empty(1)
+
+                    try:
+                        center = self.module_location.get_center()  # center of module in image
+                    except:
+                        flags.get_center = False
+
+                    try:
+                        roll = get_module_roll(
```
fallscameron01: Not sure why the module roll function got moved up here. Should be run after get_module_bounds since it depends on that. Right now it would be run on the default value of bounds.

vision/failure_flags.py
```diff
@@ -0,0 +1,60 @@
...
+class FailureFlags:
+    """
+    Abstract class used by the pipeline to track which algorithm failed
+    in the event of the program crashing
+
+    For each attribute, a value of True signals that the algorithm
+    ran to completion without any problems
+    """
+    def __init__(self):
+        pass
```
fallscameron01, line 11: While you are already making more changes, you could also add a string function to the base class that says something like "INVALID STATE".

vision/module/module_orientation.py
```diff
@@ -80,13 +80,21 @@ def get_module_roll(enclosing_region: np.ndarray) -> float:
...
+        angles = np_rectangles[:, 2]
+
+        # if the angle is absurd, it won't be taken
+        #   if the module is truly rolled at over 45 degrees, minAreaRect
+        #   should simply choose a different rectangle or axis
+        # mask = np.where(abs(angles) < 45)
+        # NOTE: the above claim was not the case. Angles with an absolute value
+        #       above 45 are common, and should probably be used to not get a 0
+        #       value in most cases
```
fallscameron01: Also should remove this long chain of comments and add a note comment about this try block.

fallscameron01 (changes requested):
Just a few small changes. Good work on the updates though.

fallscameron01 (approved):
Looks good. Nice work!

PR #249, opened 2021-04-16
Readjusted bounding boxes

### LINKED ISSUE #248: Fix text detection mapping to original image
## Description of Problem

Text detection currently does not accurately map detected text location on sliced section of image to the original image.

## Requirements

- [x] After text is detected on sliced_rotated_image, the coordinates should be mapped back to the original image.
- [x] Ensure the coordinates of the bounding boxes are correct when placed on the original image

Andrew-Gibson42:


fallscameron01 (approved):
Looks good, nice work!

PR #250, opened 2021-04-16
Feature/improve image preprocessing

### LINKED ISSUE #237: Adjust module detection preprocessing
## Description of Problem

Circle detection is finicky and requires good preprocessing. Module detection pre-processing needs to be adjusted to ensure background is not included in circle detection and holes on module can be reliable detected.

## Requirements

- [x] Adjust existing preprocessing to improve accuracy of circle detection.
- [x] Implement new preprocessing if necessary. (examples: radial blur, color filtering, increasing brightness)
- [x] Filter out circles based on other criteria. (examples: depth, distance from text, distance from edge of image)

mrouie:
Complete revamp of image preprocessing and circle post-processing

Image pre-processing:
- Increases brightness of image by a scalar value generated based on the median brightness of the image
- Switched back to Gaussian blur with a kernel of 5 and modified circle detection parameters.
- The circle detection is now more generous and has no minimum radius.
- The max radius was decreased and may need increasing as the test data set was low resolution and radius increases with higher resolution

Circle post-processing:
- Filters out circles that are either above, to the right of, to the left of, or inside the detected text using TextDetector.
- Filters out circles that are a depth of 5000 millimeters (5 meters) or more away using the depth image.
- Groups circles that are near each other into clusters and runs bucket sorting on each of the clusters to choose the best one (with the most parallels), the one that's likely the module holes.

Circle clustering:
- Circles are now clustered together to eliminate circles that are spread out and to break up and narrow down which grouping of circles is most likely the module holes
- The clustering continuously runs until a desired overall compactness of the clusters is achieved
- Clusters with less than four circles (outliers) are then removed from calculation
- Then the bucket sorting is run on each cluster to identify the best_cluster, which is the cluster with the most parallel slopes between circles detected

Other changes:
- show_circles() was removed and integrated with show_img() to make it uniform with save_img()
- The image visualization functions were also reworked to handle showing all the clusters of circles that were grouped together.
- Slight tweaking of some of the circle detection requirements such as the maximum circles allowed before in_frame or get_center fail since a lot of filtering is being applied to the circles


vision/module/location.py
```diff
@@ -51,27 +61,39 @@ def is_in_frame(self) -> bool:
...
-                self._group_slopes()
-                self.needs_recalc = False
+                num_parallels = 0
+                for cluster in self.clusters:
+                    # Attempt Slopes and Parallels with current cluster
+                    self._get_slopes(cluster)
+                    self._group_slopes(cluster)
+
+                    new_parallels = np.max(self.slope_heights)
+                    if new_parallels > num_parallels and new_parallels > MIN_PARALLELS:
```
fallscameron01: This should be
` and new_parallels >= MIN_PARALLELS:`

vision/module/location.py
```diff
@@ -51,27 +61,39 @@ def is_in_frame(self) -> bool:
...
-            self._circle_detection()
+            self._circle_detection()  # Detect circles on color image
+            self._cluster_circles()  # Group circles that are near each other
 
         # Only perform more calculations if there are a reasonable number of circles
         if (
             np.shape(self.circles)[0] <= MAX_CIRCLES
             and np.shape(self.circles)[0] >= MIN_CIRCLES
         ):  # not too little and not too many circles found
             if self.needs_recalc:
```
fallscameron01: The self.needs_recalc could be moved to the max/min circles if statement above since everything it contains everything in this block.

vision/module/location.py
```diff
@@ -51,27 +61,39 @@ def is_in_frame(self) -> bool:
...
-                # Get slopes and group parallel slopes
-                self._get_slopes()
-                self._group_slopes()
-                self.needs_recalc = False
+                num_parallels = 0
```
fallscameron01: Should rename this to `max_parallels` for clarity

vision/module/location.py
```diff
@@ -83,29 +105,38 @@ def get_center(self) -> tuple:
         -------
         tuple - (x, y) coordinates of the center of the module.
         """
-        MAX_CIRCLES = 100  # slope calculations are not performed if there are more than MAX_CIRCLES circles
+        MAX_CIRCLES = 200  # slope calculations are not performed if there are more than MAX_CIRCLES circles
         MIN_CIRCLES = 4  # minimum number of circles to perform more calculations
+        MIN_PARALLELS = (
+            8  # minimum number of main parallel required for holes to be detected
```
fallscameron01: Should change MIN_PARALLELS for this or is_in_frame for consistency. Right now is_in_frame MIN_PARALLELS is 4 and get_center MIN_PARALLELS is 8.

vision/module/location.py
```diff
@@ -83,29 +105,38 @@ def get_center(self) -> tuple:
...
-        # self._filter_circle_depth()
+            self._circle_detection()  # Detect circles on color image
+            self._cluster_circles()  # Group circles that are near each other
 
         # Only perform more calculations if there are a reasonable number of circles
         if (
             np.shape(self.circles)[0] <= MAX_CIRCLES
             and np.shape(self.circles)[0] >= MIN_CIRCLES
         ):
             if self.needs_recalc:
```
fallscameron01: Same `self.needs_recalc` comment as above.

vision/module/location.py
```diff
@@ -83,29 +105,38 @@ def get_center(self) -> tuple:
...
-                # Get Slopes and Parallels
-                self._get_slopes()
-                self._group_slopes()
-                self.needs_recalc = False
+                num_parallels = 0
+                for cluster in self.clusters:
```
fallscameron01: Since this `for cluster in self.clusters` loop is being used in both is_in_frame and get_center, it should be moved to a separate function.

vision/module/location.py
```diff
@@ -83,29 +105,38 @@ def get_center(self) -> tuple:
...
+                    # Attempt Slopes and Parallels with current cluster
+                    self._get_slopes(cluster)
+                    self._group_slopes(cluster)
```
fallscameron01: Since you are grouping the slopes multiple times, instead of `_group_slopes` assigning its results to `self.slope_heights` every time it is called, it should return the slope heights.

vision/module/location.py
```diff
@@ -126,45 +157,15 @@ def get_center(self) -> tuple:
         # or the previous center if no slope calculations were performed
```
fallscameron01: > or the previous center

This comment needs to be updated since the center is reset to (0, 1) every time the image is set.

vision/module/location.py
```diff
@@ -126,45 +157,15 @@ def get_center(self) -> tuple:
...
-                    else:
-                        new_cir = np.append(new_cir, [x, y, r])
-                    count += 1
-
-        if np.shape(new_cir)[0] != 0:
-            new_cir = new_cir.reshape((-1, 3))
-            self.circles = new_cir
-
-    def _get_hole_locations(self) -> np.ndarray:
+    def _get_hole_locations(self, circles) -> np.ndarray:
```
fallscameron01: Need to update documentation to reflect parameter change. Also need to type the parameter. Documentation should also reflect that this is being performed on a single cluster.

vision/module/location.py
```diff
@@ -187,28 +188,28 @@ def _get_hole_locations(self) -> np.ndarray:
...
+                self.holes = np.append(self.holes, circles[y])
                 hole_idx += 1
             idx += 1
 
         self.holes = self.holes.reshape((-1, 3))
         self.holes = np.unique(self.holes, axis=0)  # remove duplicates
         return self.holes
 
-    def _group_slopes(self) -> None:
+    def _group_slopes(self, circles) -> None:
```
fallscameron01: Doc/typing update here as well.

vision/module/location.py
```diff
@@ -187,28 +188,28 @@ def _get_hole_locations(self) -> np.ndarray:
...
         Bucket sort slopes to find parallels.
 
         Returns
         -------
         None
         """
-        BUCKET_MODIFIER = 0.5  # Changes how many buckets are in the range
-        NUM_CIRCLES = np.shape(self.circles)[0]  # The number of circles
+        BUCKET_MODIFIER = 2  # Changes how many buckets are in the range
+        NUM_CIRCLES = np.shape(circles)[0]  # The number of circles
```
fallscameron01: Don't need `NUM_CIRCLES` since it is not being used in this function.

vision/module/location.py
```diff
@@ -221,16 +222,17 @@ def _group_slopes(self) -> None:
...
+            + BUCKET_MODIFIER
         )
 
         # Bucket sort
         self.slope_heights, self.slope_bounds = np.histogram(
             self.slopes, self.num_buckets, (self.lower_bound, self.upper_bound)
         )
 
-    def _get_slopes(self) -> None:
+    def _get_slopes(self, circles) -> None:
```
fallscameron01: Doc/typing update

vision/module/location.py
```diff
@@ -221,16 +222,17 @@ def _group_slopes(self) -> None:
...
-            round((self.upper_bound - self.lower_bound) / bucket_width)
+        self.num_buckets: int = (
+            int(round((self.upper_bound - self.lower_bound) / bucket_width))
+            + BUCKET_MODIFIER
         )
 
         # Bucket sort
         self.slope_heights, self.slope_bounds = np.histogram(
             self.slopes, self.num_buckets, (self.lower_bound, self.upper_bound)
         )
```
fallscameron01, line 232: So here you could just return the values instead of assigning them to a member variable since this function is being called multiple times.

vision/module/location.py
```diff
@@ -249,6 +251,171 @@ def _get_slopes(self) -> None:
         # Convert slopes to degrees
         self.slopes = np.degrees(np.arctan(self.slopes))
```
fallscameron01: This should also probably be a return value for the same reasons as above.

vision/module/location.py
```diff
@@ -249,6 +251,171 @@ def _get_slopes(self) -> None:
...
+
+        Returns
+        -------
+        np.ndarray - brightened color image
+        """
+        # Get gray scale of color image
+        gray = cv2.cvtColor(src=self.img, code=cv2.COLOR_RGB2GRAY)
+
+        # Calculate magnitude of brightness based on median + offset
+        alpha_value = (128 + offset) / (np.median(gray))
```
fallscameron01: Maybe an edge case but if the median is 0 this will throw a divide by 0 error. Probably not common enough to make a change for, just noting it here.

vision/module/location.py
```diff
@@ -249,6 +251,171 @@ def _get_slopes(self) -> None:
...
+        Returns
+        -------
+        None
+        """
+        # Detect text
+        detector = TextDetector()
+        text_boxes = detector.detect_russian_word(self.img, self.depth)
+
+        # Return if text detection fails
+        if not text_boxes:
```
fallscameron01: Comment should be something like "returns all circles if text not found" (which should also be noted in the docstring)

vision/module/location.py
```diff
@@ -249,6 +251,171 @@ def _get_slopes(self) -> None:
...
+        """
+        Filters out circles that are to the left of, right of, or above text.
+
+        Returns
+        -------
+        None
+        """
+        # Detect text
+        detector = TextDetector()
+        text_boxes = detector.detect_russian_word(self.img, self.depth)
```
fallscameron01: Text detection should be done from the pipeline. Then you can have a set function to set a member variable for the bounding boxes.

Going to need to update your branch from develop since there have been pipeline changes. There shouldn't be a merge conflict though since you haven't changed the pipeline.

vision/module/location.py
```diff
@@ -249,6 +251,171 @@ def _get_slopes(self) -> None:
...
+        lr_bound = (max_x, max_y)  # lower right corner
+
+        # Filter out rectangles that are not directly below text
+        circle_filter = np.array(
+            [], dtype=np.uint16
+        )  # array of indices of circles out of bounds
+
+        i = 0
+        for x, y, _ in self.circles:
+            if y <= lr_bound[1]:  # remove if above lower bound of text
```
fallscameron01: Could just use the max/min x/y for these checks instead of making `ll_bound` and `lr_bound`. You are creating these variables and then just indexing them to get values you already have.

vision/module/location.py
```diff
@@ -249,6 +251,171 @@ def _get_slopes(self) -> None:
...
+        ll_bound = (min_x, max_y)  # lower left corner
+        lr_bound = (max_x, max_y)  # lower right corner
+
+        # Filter out rectangles that are not directly below text
+        circle_filter = np.array(
+            [], dtype=np.uint16
+        )  # array of indices of circles out of bounds
+
+        i = 0
+        for x, y, _ in self.circles:
```
fallscameron01: Could use enumerate here instead of a counter variable. Due to how Python is interpreted, counter variables are extremely slow whereas for loops can be translated into precompiled c++ code.
Would be:
```
for i, circle in enumerate(self.circles):
    x, y, _ = circle
```

relevant stackoverflow: https://stackoverflow.com/questions/522563/accessing-the-index-in-for-loops

vision/module/location.py
```diff
@@ -249,6 +251,171 @@ def _get_slopes(self) -> None:
...
+        -------
+        None
+        """
+        # Create filter of far away circles based on depth image
+        circle_filter = np.array(
+            [], dtype=np.uint16
+        )  # array of indices of distant circles
+
+        i = 0
+        for x, y, _ in self.circles:
```
fallscameron01: Same counter variable issue as above.

vision/module/location.py
```diff
@@ -249,6 +251,171 @@ def _get_slopes(self) -> None:
...
+        MAX_CLUSTERS = 30
+
+        if self.circles.shape[0] < 5:  # not enough circles for clustering
+            return self.circles
+
+        num_clusters = 2
+        compactness = COMPACT_THRESH + 1  # init compactness to enter while loop
+        while compactness > COMPACT_THRESH:  # ensure low variance within clusters
+            compactness, labels, _ = cv2.kmeans(
+                data=self.circles.astype(np.float32),  # kmeans requires float32
```
fallscameron01: Could create a new variable for the float32 circles since this is being done multiple times.

vision/module/location.py
```diff
@@ -249,6 +251,171 @@ def _get_slopes(self) -> None:
...
+                K=num_clusters,
+                bestLabels=None,
+                criteria=CLSTR_CRITERIA,
+                attempts=30,
+                flags=cv2.KMEANS_PP_CENTERS,
+            )
+            if num_clusters > MAX_CLUSTERS:
+                return self.circles
+            num_clusters += 1  # try higher K value
+        num_clusters -= 1  # decrement extra post-increment
```
fallscameron01: Can make this a bit more clear by replacing the increment/decrement with
```
if compactness > COMPACT_THRESH:
   num_clusters += 1
```
within the while loop

vision/module/location.py
```diff
@@ -257,32 +424,41 @@ def _circle_detection(self) -> np.ndarray:
...
-        blur = cv2.medianBlur(gray, 15)
+        blur = cv2.GaussianBlur(src=bright_gray, ksize=(BLUR_SIZE, BLUR_SIZE), sigmaX=0)
+        # blur = cv2.medianBlur(gray, 15)
+        # blur = cv2.bilateralFilter(src=gray, d=11, sigmaColor=75, sigmaSpace=75)
 
         # Laplacian Transform / ksize = 3 for Guassian / ksize = 1 for Median
-        laplacian = cv2.Laplacian(src=blur, ddepth=cv2.CV_8U, ksize=1)
+        laplacian = cv2.Laplacian(src=blur, ddepth=cv2.CV_8U, ksize=3)
         laplacian = np.uint8(laplacian)
+        self.laplacianimg = np.copy(laplacian)  # HACK: REMOVE THIS
```
fallscameron01: > HACK: REMOVE THIS

Indeed, remove this.

vision/module/location.py
```diff
@@ -257,32 +424,41 @@ def _circle_detection(self) -> np.ndarray:
...
-        self.circles = cv2.HoughCircles(
+        ## Hough Circle Detection ##
+        self.circles = cv2.HoughCircles(  # NOTE: Params may need partial rework, only tested on 480p data
             image=laplacian,
             method=cv2.HOUGH_GRADIENT,
             dp=1,
-            minDist=8,
-            param1=75,
-            param2=24,
+            minDist=4,
```
fallscameron01: May need to increase minimum distance, otherwise this could result in a ton of circles on higher res images.

vision/module/location.py
```diff
@@ -346,23 +495,106 @@ def set_img(self, color: np.ndarray, depth: np.ndarray) -> None:
...
 
         Returns
         -------
         None
         """
-        cv2.imshow("Module Location Image", self.img)
+        circle_img = np.copy(self.img)
+
+        if draw_circles:
+            self.get_center()
```
fallscameron01: Should this be calling get_center? You would probably want to have the user call get_center and then this function.

vision/module/location.py
```diff
@@ -346,23 +495,106 @@ def set_img(self, color: np.ndarray, depth: np.ndarray) -> None:
...
+                    cv2.rectangle(
+                        img=circle_img,
+                        pt1=(x - 1, y - 1),
+                        pt2=(x + 1, y + 1),
+                        color=color,
+                        thickness=-1,
+                    )
+
+        if draw_center:
+            self.get_center()
```
fallscameron01: Same thing here.

fallscameron01 (changes requested):
Really good work on this. Lots of good additions and modifications.
For things to fix, there are a lot of smaller changes/adjustments throughout, as well as documentation fixes. Big thing is moving the text detection to the pipeline instead of calling it within the location class. You should just be able to move the text detection to the pipeline and then create a new member variable and setter in the location class to store the bounding boxes. You'll need to update your branch from develop since there have been pipeline changes.
Overall really good job though!

vision/module/location.py
```diff
@@ -83,171 +87,340 @@ def get_center(self) -> tuple:
         -------
         tuple - (x, y) coordinates of the center of the module.
         """
-        MAX_CIRCLES = 100  # slope calculations are not performed if there are more than MAX_CIRCLES circles
+        MAX_CIRCLES = 200  # slope calculations are not performed if there are more than MAX_CIRCLES circles
         MIN_CIRCLES = 4  # minimum number of circles to perform more calculations
 
+        # Coordinates of the center of the front face of the module
+        self.center = np.arange(0, 2)  # default center value (0, 1)
```
fallscameron01: This is unnecessary since the default value is set whenever `set_img` is called.

vision/module/location.py
```diff
@@ -83,171 +87,340 @@ def get_center(self) -> tuple:
...
+            num_holes += 1
 
-            self.center[0] = x_total // num_holes
-            self.center[1] = y_total // num_holes
+        self.center[0] = x_total // num_holes
+        self.center[1] = y_total // num_holes
 
         # Returns either the center in the current image
-        # or the previous center if no slope calculations were performed
+        # or the center (0, 1) if no center was found
```
fallscameron01, line 124: Should note that (0, 1) is the default value in the docstring.

vision/module/location.py
```diff
@@ -346,64 +500,146 @@ def set_img(self, color: np.ndarray, depth: np.ndarray) -> None:
...
+
         self.center = np.arange(2)
-        self.needs_recalc = True
+        self.holes = np.zeros((4, 3), dtype=np.uint16)
 
-    ## Visualization Functions
+        self.needs_recalc = True
 
-    def show_img(self) -> None:
+    def set_text(self, text_boxes: BoundingBox) -> None:
```
fallscameron01: > text_boxes: BoundingBox

`text_boxes` is a list of bounding boxes.

vision/module/location.py
```diff
@@ -83,171 +87,340 @@ def get_center(self) -> tuple:
...
+                self.holes = np.append(self.holes, self.best_cluster[y])
                 hole_idx += 1
             idx += 1
 
         self.holes = self.holes.reshape((-1, 3))
         self.holes = np.unique(self.holes, axis=0)  # remove duplicates
         return self.holes
 
-    def _group_slopes(self) -> None:
+    def _group_slopes(self, slopes: np.ndarray) -> np.ndarray:
```
fallscameron01: >  -> np.ndarray:

Return type should be tuple.

vision/obstacle/obstacle_tracker.py
```diff
@@ -94,7 +94,7 @@ def update(self, new_obstacle_boxes: list) -> None:
         new_obstacles = np.array([])
 
         # transform the list of bounding boxes into an array of Obstacles
-        for i in np.arange(len(new_obstacle_boxes)):
+        for i in np.arange(np.size(new_obstacle_boxes)):
```
fallscameron01: This should be len instead of size. size propagates down the dimensions of the array so it will get a different value than len. Changed to len by #247

vision/pipeline.py
```diff
@@ -112,25 +113,42 @@ def run(self, prev_state):
...
                 flags.set_img = False
 
+            if flags.set_img:
+                try:
+                    text_bboxes = self.text_detector.detect_russian_word(
+                        color_image, depth_image
+                    )
+                except:
+                    text_bboxes = []
+                    flags.detect_russian_word = False
```
fallscameron01, line 143: How the flags work is that there is a different flag object for each state, so you should add this flag to the `ModuleDetectionFlags` class in `failure_flags.py`

vision/pipeline.py
```diff
@@ -112,25 +113,42 @@ def run(self, prev_state):
...
+
+            if flags.detect_russian_word:
+                try:
+                    self.module_location.set_text(text_bboxes)
+                except:
+                    flags.set_text = False
```
fallscameron01, line 149: Same thing for this flag. If you don't add the flags to the actual class, it won't exist when you check it below since it is created locally in this if block.

vision/module/location.py
```diff
@@ -346,64 +500,146 @@ def set_img(self, color: np.ndarray, depth: np.ndarray) -> None:
...
         self.img = color
+
+        # Re-initialize all important values to zero/empty
         self.circles = np.array([])
+        self.clusters = []
+        self.best_cluster = np.array([])
+
+        self.best_slopes = np.array([])
+        self.best_grouped_slopes = (np.array([0]), np.array([0]), 0)
+
```
fallscameron01, line 512: Should `set_img` also be setting `self.text_boxes` to an empty array?

vision/pipeline.py
```diff
@@ -112,25 +113,42 @@ def run(self, prev_state):
...
-                self.module_location.setImg(color_image, depth_image)
+                self.module_location.set_img(color_image, depth_image)
             except:
                 flags.set_img = False
 
+            if flags.set_img:
+                try:
+                    text_bboxes = self.text_detector.detect_russian_word(
+                        color_image, depth_image
+                    )
```
fallscameron01, line 141: Since you are now getting bounding boxes for text in this state you should append these to bboxes as well. That way, there will be both text and module bounding boxes for flight to work with.

vision/pipeline.py
```diff
@@ -149,7 +167,9 @@ def run(self, prev_state):
                     bounds = np.empty(1)
 
                     try:
-                        center = self.module_location.get_center()  # center of module in image
+                        center = (
+                            self.module_location.get_center()
+                        )  # center of module in image]
```
fallscameron01, line 171: > center of module in image]

random bracket

fallscameron01 (changes requested):
Some more small changes to make throughout. Good work on fixes the issues from the last review!

vision/obstacle/obstacle_tracker.py
mrouie: I'm not actually sure when or how this got changed but it was unintentional. I noticed it in the merge conflict and didn't think much of it. Thanks.

vision/pipeline.py
```diff
@@ -112,25 +113,42 @@ def run(self, prev_state):
...
+                self.module_location.set_img(color_image, depth_image)
             except:
                 flags.set_img = False
 
+            if flags.set_img:
+                try:
+                    text_bboxes = self.text_detector.detect_russian_word(
+                        color_image, depth_image
+                    )
+                    bboxes.append(text_bboxes)
```
fallscameron01: Since text_bboxes is a list of bounding boxes, this would append a list to bboxes instead of appending the bounding boxes. Can just set `bboxes = text_bboxes` or use `bboxes.extend(text_bboxes)`

fallscameron01 (changes requested):
One small change and then you're good.

fallscameron01 (approved):
Nice job! Really good work on this. You worked really hard and made a lot of good changes!

PR #253, opened 2021-04-21
async pipe

### LINKED ISSUE #244: Asynchronous pipeline
## Description of Problem

Currently, the vision code is designed to run synchronously, which may be causing issues with flight code (asynchronous). Therefore, an asynchronous version of the pipeline should be implemented to narrow down list of issues discovered at flight test.

## Requirements

- [x] Implement pipeline asynchronously.
- [x] Ensure pipeline runs do not overlap. (should wait for previous run to finish)

mat-px:
-Made run function in Pipeline "asynchronous"
-Made init_vision "asynchronous"
-Created task init_vision in main
-Added async for loop function

### DIFF
--- a/vision/pipeline.py
+++ b/vision/pipeline.py
@@ -8,10 +8,6 @@ import numpy as np
 import asyncio
 import warnings
 
-warnings.filterwarnings("ignore")
-# import time
-# start_time = time.time()
-
 parent_dir = os.path.dirname(os.path.abspath(__file__))
 gparent_dir = os.path.dirname(parent_dir)
 ggparent_dir = os.path.dirname(gparent_dir)
@@ -43,6 +39,17 @@ from vision.failure_flags import ObstacleDetectionFlags
 from vision.failure_flags import TextDetectionFlags
 from vision.failure_flags import ModuleDetectionFlags
 
+async def arange(count: int) -> int:
+    """
+    Asynchronous loop to iterate over a count
+
+    Parameters
+    -------------
+    count: integer
+        Amount of times to asynchronously iterate
+    """
+    for i in range(count):
+        yield(i)
 
 class Pipeline:
     """
@@ -217,10 +224,6 @@ class Pipeline:
                                             :,
                                         ]
                                     )  # roll of module
-                                except:
-                                    flags.get_module_roll = False
-
-                                if flags.get_module_roll:
                                     # construct boundingbox for the module
                                     box = BoundingBox(bounds, ObjectType.MODULE)
                                     box.module_depth = depth  # float
@@ -228,42 +231,40 @@ class Pipeline:
                                         roll,
                                     )  # x, y, z tilt
                                     bboxes.append(box)
+                                except:
+                                    flags.get_module_roll = False
 
         else:
             pass  # raise AttributeError(f"Unrecognized state: {state}")
 
+        time = datetime.datetime.now()
+
+        # You need to index vision_flags to see output, "self.vision_flags.get()[1]"
         self.vision_flags.put(
-            (datetime.datetime.now(), flags), self.PUT_TIMEOUT
+            (time, flags), self.PUT_TIMEOUT
         )
         await asyncio.sleep(0.01)
         ##
         self.vision_communication.put(
-            (datetime.datetime.now(), bboxes), self.PUT_TIMEOUT
+            (time, bboxes), self.PUT_TIMEOUT
         )
         await asyncio.sleep(0.01)
         # uncomment to visualize blobs
         # from vision.common.blob_plotter import plot_blobs
         # plot_blobs(self.obstacle_finder.keypoints, color_image)
 
-        # print(self.vision_flags.get()[1])
-
         return state
 
 
-async def init_vision(vision_comm, flight_comm, video, runtime=100):
+async def init_vision(vision_comm, flight_comm, camera, state, runtime=100):
     """
     Alex, call this function - not run.
     """
-    pipeline = Pipeline(vision_comm, flight_comm, video)
+    pipeline = Pipeline(vision_comm, flight_comm, camera)
 
-    state = "module_detection"
-
+    warnings.filterwarnings("ignore")
     async for _ in arange(runtime):
         await pipeline.run(state)
-
-async def arange(count):
-    for i in range(count):
-        yield(i)
 
 if __name__ == "__main__":
     from vision.camera.bag_file import BagFile
@@ -278,9 +279,10 @@ if __name__ == "__main__":
     video_file = sys.argv[1]
     video = BagFile(100, 100, 60, video_file)
 
-    loop = asyncio.get_event_loop()
+    state = "module_detection"
 
-    asyncio.run(init_vision(vision_comm, flight_comm, video))
+    loop = asyncio.get_event_loop()
+    asyncio.run(init_vision(vision_comm, flight_comm, video, state))
 
     from time import sleep
 
@@ -288,4 +290,3 @@ if __name__ == "__main__":
 
     vision_comm.close()
     flight_comm.close()
-    # print("--- %s seconds ---" % (time.time() - start_time))
--- a/vision/pipeline.py
+++ b/vision/pipeline.py
@@ -5,6 +5,12 @@ Takes information from the camera and gives it to vision
 import os
 import sys
 import numpy as np
+import asyncio
+import warnings
+
+warnings.filterwarnings("ignore")
+# import time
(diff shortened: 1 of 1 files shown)


vision/pipeline.py
```diff
@@ -5,6 +5,12 @@
 import os
 import sys
 import numpy as np
+import asyncio
+import warnings
+
+warnings.filterwarnings("ignore")
```
fallscameron01: Should put the ignore warnings in the pipeline's init function.

vision/pipeline.py
```diff
@@ -5,6 +5,12 @@
...
+# import time
+# start_time = time.time()
```
fallscameron01: Need to remove these comments.

vision/pipeline.py
```diff
@@ -222,37 +227,43 @@ def run(self, prev_state):
...
         self.vision_communication.put(
             (datetime.datetime.now(), bboxes), self.PUT_TIMEOUT
         )
-
+        await asyncio.sleep(0.01)
         # uncomment to visualize blobs
         # from vision.common.blob_plotter import plot_blobs
         # plot_blobs(self.obstacle_finder.keypoints, color_image)
 
+        # print(self.vision_flags.get()[1])
```
fallscameron01: And this comment here.

vision/pipeline.py
```diff
@@ -222,37 +227,43 @@ def run(self, prev_state):
...
 
-def init_vision(vision_comm, flight_comm, video, runtime=100):
+async def init_vision(vision_comm, flight_comm, video, runtime=100):
     """
     Alex, call this function - not run.
     """
     pipeline = Pipeline(vision_comm, flight_comm, video)
 
-    state = "start"
+    state = "module_detection"
```
fallscameron01: You should add a state parameter for this function so the state can be set externally to it. Also, the `video` parameter should be renamed to `camera`.

vision/pipeline.py
```diff
@@ -222,37 +227,43 @@ def run(self, prev_state):
...
 
-    for _ in range(runtime):
-        state = pipeline.run(state)
+    async for _ in arange(runtime):
+        await pipeline.run(state)
 
+async def arange(count):
+    for i in range(count):
+        yield(i)
```
fallscameron01: Should move the arange definition to the top of the file under the imports.

vision/pipeline.py
```diff
@@ -222,37 +227,43 @@ def run(self, prev_state):
                                     box.orientation = orientation + (
```
fallscameron01: This is unrelated to the asynchronous stuff but could you remove this `if flags.get_module_roll` statement. This part in the if statement should just be under the `if flags.get_module_bounds` statement.

vision/pipeline.py
```diff
@@ -79,13 +85,13 @@ def __init__(self, vision_communication, flight_communication, camera):
 
         self.module_location = ModuleLocation()
 
-        self.vision_flags = []
+        self.vision_flags = Queue()
```
fallscameron01, line 97: As a side note, should add a note about how the tuple returned from `vision_flags.get()` needs to be indexed to see the output from the vision flags rather than the object reference.

vision/pipeline.py
```diff
@@ -222,37 +227,43 @@ def run(self, prev_state):
...
 
-        self.vision_flags.append(flags)
-
+        self.vision_flags.put(
+            (datetime.datetime.now(), flags), self.PUT_TIMEOUT
+        )
+        await asyncio.sleep(0.01)
         ##
         self.vision_communication.put(
             (datetime.datetime.now(), bboxes), self.PUT_TIMEOUT
```
fallscameron01: Should add a variable above these two put statements where you get the datetime so that the two queues have the same time for the same frame.

vision/pipeline.py
```diff
@@ -222,37 +227,43 @@ def run(self, prev_state):
...
 
-    state = "start"
+    state = "module_detection"
 
-    for _ in range(runtime):
-        state = pipeline.run(state)
+    async for _ in arange(runtime):
+        await pipeline.run(state)
 
+async def arange(count):
```
fallscameron01: Also should type the parameters and return types for `arange`, as well as the rest of the functions in this file. `arange` also needs a docstring.
By typing I mean this:
`async def arange(count: int) -> int:`

vision/pipeline.py
```diff
@@ -267,11 +278,14 @@ def init_vision(vision_comm, flight_comm, video, runtime=100):
...
+
+    asyncio.run(init_vision(vision_comm, flight_comm, video))
 
     from time import sleep
 
     sleep(2)  # allow queue to close
 
     vision_comm.close()
     flight_comm.close()
+    # print("--- %s seconds ---" % (time.time() - start_time))
```
fallscameron01: Also, just for future reference, you should put your start time at the top of main rather than the top of the file so that you aren't including time waiting for imports.

fallscameron01 (changes requested):
There's a few small changes to make throughout including some formatting and documentation updates, as well as a update to `init_vision` and a general pipeline update.

vision/pipeline.py
```diff
@@ -212,47 +224,47 @@ def run(self, prev_state):
                                             :,
                                         ]
                                     )  # roll of module
-                                except:
-                                    flags.get_module_roll = False
```
fallscameron01, line 216: Shouldn't have moved this except section. The stuff below shouldn't be in the try block.

vision/pipeline.py
```diff
@@ -212,47 +224,47 @@ def run(self, prev_state):
...
-
-                                if flags.get_module_roll:
                                     # construct boundingbox for the module
                                     box = BoundingBox(bounds, ObjectType.MODULE)
                                     box.module_depth = depth  # float
                                     box.orientation = orientation + (
```
fallscameron01: Also need to unindent this so it is indented once from `if flags.get_module_bounds:`. These bounding box lines shouldn't be in the try block.

vision/pipeline.py
```diff
@@ -212,47 +224,47 @@ def run(self, prev_state):
...
+                                except:
+                                    flags.get_module_roll = False
 
         else:
             pass  # raise AttributeError(f"Unrecognized state: {state}")
 
-        self.vision_flags.append(flags)
+        time = datetime.datetime.now()
 
+        # You need to index vision_flags to see output, "self.vision_flags.get()[1]"
```
fallscameron01, line 244: Something I though of: If you change the `__str__()` functions in the flag classes to `__repr__()`, it might fix this so that the tuple doesn't need to be indexed. Should be as simple as changing the function names if you could test this. Classes are in `failure_flags.py`

vision/pipeline.py
```diff
@@ -212,47 +224,47 @@ def run(self, prev_state):
...
     """
-    pipeline = Pipeline(vision_comm, flight_comm, video)
-
-    state = "start"
-
-    for _ in range(runtime):
-        state = pipeline.run(state)
+    pipeline = Pipeline(vision_comm, flight_comm, camera)
 
+    warnings.filterwarnings("ignore")
```
fallscameron01: Oh, sorry, by init I mean the filter warnings should go in the `pipeline`'s member function `__init__()`

vision/pipeline.py
```diff
@@ -212,47 +224,47 @@ def run(self, prev_state):
...
+        await asyncio.sleep(0.01)
         # uncomment to visualize blobs
         # from vision.common.blob_plotter import plot_blobs
         # plot_blobs(self.obstacle_finder.keypoints, color_image)
 
         return state
 
 
-def init_vision(vision_comm, flight_comm, video, runtime=100):
+async def init_vision(vision_comm, flight_comm, camera, state, runtime=100):
```
fallscameron01: Also should type the parameters and returns for any functions in this file. Would be:
`async def init_vision(vision_comm: Queue, flight_comm: Queue, camera: Camera, state: str, runtime: int = 100) -> None:`

Should do this for the rest of the functions in this file.

vision/pipeline.py
```diff
@@ -212,47 +224,47 @@ def run(self, prev_state):
...
     """
     Alex, call this function - not run.
     """
```
fallscameron01, line 277: Also need to update this docstring.

fallscameron01 (changes requested):
Couple of changes that weren't done correctly. Also should type the parameters and returns for all the function in this file. And one other change I thought of that might fix the issue with having to index the flags.

fallscameron01 (approved):
Good work! Ready for merge.

PR #254, opened 2021-04-24
Improve overall text detection reliability and update text benchmark

### LINKED ISSUE #251: Ensure text in min area rectangle when mast in view
## Description of Problem

When the mast is in view, the minimum area rectangle calculated by the text detector may not contain the text. This results in tesseract not finding the text or getting false positives on dataset with mast in view. Need to update to make sure that rectangle always contains the text.

## Requirements

- [x] Ensure text always in min area rectangle
- [x] Need to update mask so more of mast ends up checked for contours.

mrouie:
In TextDetector.\_\_init\_\_() from detect_words.py,
The stored member variables of the TextDetector class have been updated to separate the two words on the communication module for discrete comparison  with the detected text. 
A new member variable has also been added to store the tesseract data generated. This is done for easy access of the data generated by tesseract for benchmarking and unit testing programs.

In TextDetector.detect_russian_word() from detect_words.py,
The text detection success parameters have been tweaked to only return BoundingBoxes that fulfil the requisite of having at least one whole word of the two on the communication module exist in the tesseract text data.

In TextDetector._get_rotated_min_area_rect() from detect_words.py,
The mask generated over the color image used to find the minimum area rectangle of the text has been updated to work with a a far wider range of brightness levels as it now uses a weighted ratio between the blue & green and blue & red channels.

In TextDetector._get_rotated_min_area_rect() from detect_words.py,
The function now consistently correctly rotates the detected minimum area rectangle. 
Previously, the text detection would not work as intended on a Linux operating system due to OpenCV returning different a shifted value range for the cv2.minAreaRect() function. This is now corrected for Linux, but there is a commented out portion left for Windows developers and testers.

In BenchTextAccuracy.bench_accuracy() from bench_text.py,
Errors in the benchmark were now resolved.
The benchmark now outputs the text-detection-generated BoundingBoxes and full tesseract data to a file for each image file the benchmark processes to a "results" folder in the root directory of the file run.
Additionally, like before, the benchmark also can output marked image files of the detected text boxes to a "marked_images" folder in the root directory of the file run

### DIFF
--- a/vision/text/detect_words.py
+++ b/vision/text/detect_words.py
@@ -21,7 +21,8 @@ class TextDetector:
     """
 
     def __init__(self):
-        self.text = "модулииртибот"
+        self.text = np.array(["модули", "иртибот"])
+        self.tessdata = dict()
 
     def detect_russian_word(
         self, color_image: np.ndarray, depth_image: np.ndarray
@@ -41,34 +42,36 @@ class TextDetector:
         A list of bounding box objects that contain desired text
             Note that there may be different bounding boxes for each word
         """
-        sliced_rotated_image, x_ul, y_ul = self._get_rotated_min_area_rect(color_image, depth_image)
+        sliced_rotated_image, x_ul, y_ul = self._get_rotated_min_area_rect(
+            color_image, depth_image
+        )
 
         if len(sliced_rotated_image) == 0:
             return []
 
-        tessdata = pytesseract.image_to_data(
+        self.tessdata = pytesseract.image_to_data(
             sliced_rotated_image, output_type=pytesseract.Output.DICT, lang="uzb_cyrl"
         )
 
-        n_boxes = len(tessdata["level"])
+        detected_words = self.tessdata["text"]
         box_obs = []
-        contents = tessdata["text"]
-        for i in range(n_boxes):
-            if not contents[i]:
-                continue
-            else:
-                for j in contents[i]:
-                    if j in self.text:
-                        (x, y, w, h) = (
-                            tessdata["left"][i] + x_ul,
-                            tessdata["top"][i] + y_ul,
-                            tessdata["width"][i],
-                            tessdata["height"][i],
-                        )
-                        verts = [(x, y), (x + w, y), (x, y + h), (x + w, y + h)]
-                        box = BoundingBox(verts, ObjectType("text"))
-                        box_obs.append(box)
-                        break
+
+        for i, det_word in enumerate(detected_words):
+            if not det_word:
+                continue  # ignore empty strings
+
+            # if any of the words are in the detected words, add BoundingBox to final set
+            # This is done to allow cases where extra characters around the words are detected
+            if np.any([(rus_word in det_word.lower()) for rus_word in self.text]):
+                (x, y, w, h) = (
+                    self.tessdata["left"][i] + x_ul,
+                    self.tessdata["top"][i] + y_ul,
+                    self.tessdata["width"][i],
+                    self.tessdata["height"][i],
+                )
+                verts = [(x, y), (x + w, y), (x, y + h), (x + w, y + h)]
+                box = BoundingBox(verts, ObjectType("text"))
+                box_obs.append(box)
 
         return box_obs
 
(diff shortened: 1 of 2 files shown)


vision/text/detect_words.py
```diff
@@ -21,7 +21,8 @@ class TextDetector:
     """
 
     def __init__(self):
-        self.text = "модулииртибот"
+        self.text = np.array(["модули", "иртибот"])
+        self.tessdata = dict()
```
fallscameron01: Best practice would be to initialize a dictionary like this:
`self.tessdata: dict = {}`
to emphasize that it is empty while also displaying the type.

vision/text/detect_words.py
```diff
@@ -90,6 +93,9 @@ def _get_rotated_min_area_rect(
         min area rect: np.ndarray
             rotated ndarray of the largest contour
```
fallscameron01, line 98: Return type needs to be updated since this function returns a 3-tuple.

vision/text/detect_words.py
```diff
@@ -99,7 +105,12 @@ def _get_rotated_min_area_rect(
             blur_image[:, :, 2],
         )  # separate channels
 
-        isblue = np.logical_and((b_image - r_image > 120), (b_image - g_image > 20))
+        r_image = np.where(r_image == 0, 1, r_image)  # prevent division by zero
+        g_image = np.where(r_image == 0, 1, r_image)  # prevent division by zero
+
+        isblue = np.logical_and(
+            (b_image / r_image > RED_WT), (b_image / g_image > GREEN_WT)
```
fallscameron01: It looks like this would count near-black colors in the image as blue. For example if the RGB pixel value is (1, 1, 2), it would be counted as blue even though it is basically black. Should add a minimum blue threshold of 20 or so to this `and` statement, so that there would have to be a minimum blue value to be counted as blue.

vision/text/detect_words.py
```diff
@@ -99,7 +105,12 @@ def _get_rotated_min_area_rect(
             blur_image[:, :, 2],
         )  # separate channels
 
-        isblue = np.logical_and((b_image - r_image > 120), (b_image - g_image > 20))
+        r_image = np.where(r_image == 0, 1, r_image)  # prevent division by zero
+        g_image = np.where(r_image == 0, 1, r_image)  # prevent division by zero
+
+        isblue = np.logical_and(
```
fallscameron01: Also should rename this variable to something like `blue_mask` for clarity.

vision/text/detect_words.py
```diff
@@ -119,7 +130,7 @@ def _get_rotated_min_area_rect(
         )
```
fallscameron01, line 149: A couple of things unrelated to this PR but I want to get the code cleaned up since you are already working on it.
Can't add my comment on the exact line number but on line 118 there is a magic number `depth_image < 8000` that should be a constant.
Also the parameters for cv2 functions in this block should be keyword-specified for clarity on Canny, GuassianBlur, and findContours.

vision/text/detect_words.py
```diff
@@ -119,7 +130,7 @@ def _get_rotated_min_area_rect(
...
 
         if hierarchy is None:
-            return []
+            return np.array([]), None, None
```
fallscameron01: The None returns here should be 0 to keep the data type consistent.

vision/text/detect_words.py
```diff
@@ -41,34 +42,36 @@ def detect_russian_word(
...
-        sliced_rotated_image, x_ul, y_ul = self._get_rotated_min_area_rect(color_image, depth_image)
+        sliced_rotated_image, x_ul, y_ul = self._get_rotated_min_area_rect(
+            color_image, depth_image
+        )
 
         if len(sliced_rotated_image) == 0:
             return []
 
-        tessdata = pytesseract.image_to_data(
+        self.tessdata = pytesseract.image_to_data(
```
fallscameron01, line 54: `self.tessdata` should be set to empty at the start of this function so that if the function returns at `if len(sliced_rotated_image) == 0:` the tessdata isn't from the previous image.

fallscameron01 (changes requested):
Lots of good changes to the mask and the text parsing. A few smaller changes for cleanup and clarity. One thing is that your blue mask would currently count the RGB pixel (1, 1, 2) as blue, even though it is basically black, so there should be a minimum blue threshold added.

fallscameron01 (approved):
Nice job cleaning things up! Ready to merge.

PR #260, opened 2021-04-28
Rework text detection with rotated bounding boxes and rework module location text filter accordingly

### LINKED ISSUE #257: Module location handles single text bounding box
## Description of Problem

Right now, filtering in module location assumes that the only cases will be 0 or 2 bounding boxes. Need to handle other cases, especially the case where only one word was found.

## Requirements

- [x] Adjust to handle cases where one word was found.
- [x] Handle cases where more than two boxes found, most likely need to just return full circle list. This case would be extremely rare, but should be accounted for anyway.

### LINKED ISSUE #258: Remap text angle
## Description of Problem

Right now, remapping text from the min area rectangle to the original image does not take the angle of the text into account to find the other vertices, resulting in the text always being horizontal. Not a problem for the text detection itself, but could throw off module location circle filtering.

## Requirements

- [x] Take angle of min area rectangle into account for mapping
- [x] Take angle of text into account for mapping (if necessary)
- [x] Map all 4 text vertices so bounding box can be skewed relative to image. (Should still be rectangle but may be rotated relative to image)

mrouie:
- TextDetection now returns bounding boxes that have been rotated to account for the angle that the text is tilted at. 
New text boxes are represented by green tilted rectangles in reference image 1.
- TextDetection._get_rotated_min_area_rect() was updated with a conditional workaround for a potential bug in the opencv library
- ModuleLocation's text filtering now uses vector cross products to check if text is below. Cross products between bounding vectors (dark cyan) and vectors to circles (red/white) indicate the relative location of the circle in question. The sign of the cross product determines which side of the text the circle is on. In this case the circles outside the dark cyan bounds (red vectors) were filtered out while the circles inside the cyan bounds (being directly below the text) were kept (white vectors).
- Reference image 1 (from RealSense bag file 1 from the 2/27/21 flight test):
bag file 1 - all vectors from both text bounds to all circles
- Additionally, in ModuleLocation, the text filtering now accounts for if there's only one text box detected by adding a padded margin around the text box of a thickness equal to half the width of the text box (See Reference Image 2 and 3
- Reference image 2 - with padded text bounding (from RealSense bag file 1 from the 2/27/21 flight test):
text filtering - only one word detected
Reference image 2 - with vectors (from RealSense bag file 1 from the 2/27/21 flight test):
text filtering - only one word detected - with arrows
- The module location benchmark was updated to utilize text filtering
- The benchmark driver was updated with a quick bug fix that caused an error if an algorithm had crashed

### DIFF
--- a/vision/module/location.py
+++ b/vision/module/location.py
@@ -297,38 +297,57 @@ class ModuleLocation:
 
     def _filter_text_circles(self) -> np.ndarray:
         """
-        Filters out circles that are to the left of, right of, or above text.
+        Filters out circles that are to the left of, right of, or above set text.
 
         Returns
         -------
         np.ndarray - filtered array of circles
         """
-        # Return if no text boxes
+        # Can't filter using text that doesn't exist
         if not self.text_boxes:
             return self.circles
-
-        # Convert BoundingBoxes to list of values per axis
-        vertices = np.array(
-            [vertex for bbox in self.text_boxes for vertex in bbox.vertices]
-        )  # unravel into list of vertices
-        axes_vals = vertices.transpose()  # axis_vals[0] = x, axis_vals[1] = y
-
-        # Find max and min values for each axis
-        x_ubound = np.max(axes_vals[0])  # horizontal upper bound of text
-        x_lbound = np.min(axes_vals[0])  # horizontal lower bound of text
-        y_lbound = np.max(axes_vals[1])  # vertical lower bound of text
 
         # Filter out rectangles that are not directly below text
         circle_filter = np.array(
             [], dtype=np.uint16
         )  # array of indices of circles out of bounds
 
+        # get text bounding extreme points of rotated bounding boxes
+        ll_x, ll_y = self.text_boxes[0].vertices[0]  # lower left point
+        # ul_x, ul_y = self.text_boxes[0].vertices[1] #NOTE: not needed
+        # ur_x, ur_y = self.text_boxes[-1].vertices[2] #NOTE: not needed
+        lr_x, lr_y = self.text_boxes[-1].vertices[3]  # lower right point
+
+        rot_neg90 = np.array([[0, 1], [-1, 0]])  # matrix to rotate a vector -90 degrees
+
+        # bounding vectors housing the allowed region
+        lr_vect = (lr_x - ll_x, lr_y - ll_y)  # vector from ll bound to lr bound of text
+        ll_vect = (lr_x - ll_x, lr_y - ll_y)  # vector from lr bound to ll bound of text
+
+        # horizonal bounding vectors perpendicular to lower bound vectors (pointed downwards)
+        left_vect = tuple(np.dot(lr_vect, rot_neg90).astype(int))  # left bound vector
+        right_vect = tuple(np.dot(ll_vect, rot_neg90).astype(int))  # right bound vector
+
+        # create a vector from the lower-left/lower-right corner of text to all detected circles
+        # the cross product tells us what side of the bounding vectors the circle is on
         for i, (x, y, _) in enumerate(self.circles):
-            if y <= y_lbound:  # remove if above lower bound of text
-                circle_filter = np.append(circle_filter, i)
-            elif (
-                x <= x_lbound or x >= x_ubound
-            ):  # remove if horizontally outside text bounds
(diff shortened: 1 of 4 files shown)


vision/module/location.py
```diff
@@ -297,38 +297,57 @@ def _increase_brightness(self, offset: int) -> np.ndarray:
...
-
         # Filter out rectangles that are not directly below text
         circle_filter = np.array(
             [], dtype=np.uint16
         )  # array of indices of circles out of bounds
 
+        # get text bounding extreme points of rotated bounding boxes
+        ll_x, ll_y = self.text_boxes[0].vertices[0]  # lower left point
+        # ul_x, ul_y = self.text_boxes[0].vertices[1] #NOTE: not needed
+        # ur_x, ur_y = self.text_boxes[-1].vertices[2] #NOTE: not needed
```
fallscameron01: If these aren't needed, they should be removed.

vision/module/location.py
```diff
@@ -297,38 +297,57 @@ def _increase_brightness(self, offset: int) -> np.ndarray:
...
+        lr_x, lr_y = self.text_boxes[-1].vertices[3]  # lower right point
```
fallscameron01, line 324: Should handle cases where there is only one text box. How it is now, if we only have the right text box, it will filter out circles to the left of it, which can include the module circles.

vision/module/location.py
```diff
@@ -297,38 +297,57 @@ def _increase_brightness(self, offset: int) -> np.ndarray:
...
+        Filters out circles that are to the left of, right of, or above set text.
 
         Returns
         -------
         np.ndarray - filtered array of circles
         """
-        # Return if no text boxes
+        # Can't filter using text that doesn't exist
         if not self.text_boxes:
             return self.circles
```
fallscameron01, line 315: Also should handle cases with >=3 text boxes by returning here. This case is extremely unlikely with your updates to text detection, but it should be handled anyway.

vision/module/location.py
```diff
@@ -297,38 +297,57 @@ def _increase_brightness(self, offset: int) -> np.ndarray:
...
+
+            lower_cross = np.cross(lr_vect, ll_to_circ)  # <0 = above lower bound
+            if lower_cross <= 0:
+                needs_removal = True
+            left_cross = np.cross(left_vect, ll_to_circ)  # >0 = outside left bound
+            if left_cross >= 0:
+                needs_removal = True
+            right_cross = np.cross(right_vect, lr_to_circ)  # <0 = outside right bound
+            if right_cross <= 0:
+                needs_removal = True
```
fallscameron01, line 361: Really good idea taking the cross products to determine location.

vision/text/detect_words.py
```diff
@@ -58,6 +61,9 @@ def detect_russian_word(
         detected_words = self.tessdata["text"]
         box_obs = []
 
+        rows, columns = color_image.shape[0], color_image.shape[1]
```
fallscameron01: Instead of indexing, you can do this:
```suggestion
        rows, columns, _ = color_image.shape
```

vision/text/detect_words.py
```diff
@@ -71,7 +77,24 @@ def detect_russian_word(
...
+
+                # text detection is performed on a rotated image
+                # boxes need to be rotated back to their original position
+                rotated_verts = [ll_pt, ul_pt, ur_pt, lr_pt]
+
+                # rotate box back to original point on color image
+                rot_mat = cv2.getRotationMatrix2D(
+                    center=(x_ul, y_ul), angle=theta, scale=1
+                )
+                verts = cv2.transform(src=np.array([rotated_verts]), m=rot_mat)[0]
```
fallscameron01: > src=np.array([rotated_verts])

Instead of casting `rotated_verts` in the middle of this function call, you can do the cast when you initialize `rotated_verts` since it is not being used anywhere else.

fallscameron01 (changes requested):
A couple of smaller changes, but the main thing is that, for circle filtering, you should be handling cases with 1 text box or >=3 text boxes.

vision/module/location.py
mrouie: Thanks

fallscameron01 (approved):
Nice work. Ready for merge.

PR #261, opened 2021-04-29
Resolve mean of empty slice and ragged array numpy warnings

### LINKED ISSUE #234: Track down and fix numpy warnings
## Description of Problem

Algorithms frequently throw numpy errors. Need to find out where and why this is happening and fix.
Run file from command line with `-W error` to make warnings fatal or use `import warnings` `warnings.simplefilter("error")` within the program. Either of these will make warnings cause crashes and give a line number.

## Requirements

- [x] Track down sources of warnings.
- [x] Fix warnings. Should be able to run benchmarks and pipeline without triggering warnings.

mrouie:
- Resolved mean of empty slice numpy warnings caused by module depth and module roll functions
Mean of empty slice warning occurs when numpy attempts numpy.mean on an empty numpy array
- Resolved numpy ragged array warning caused by module roll's casting of the rectangles, being a ragged 2D python list, to numpy array without labeling the data type as "object"

### DIFF
--- a/vision/module/get_module_depth.py
+++ b/vision/module/get_module_depth.py
@@ -57,9 +57,11 @@ def get_module_depth(depth_image: np.ndarray, coordinates: tuple) -> float:
     # Gets rid of 0 depth values
     depth_values_in_radius = depth_values_in_radius[depth_values_in_radius != 0]
 
-    avg = np.mean(depth_values_in_radius)
-    if np.isnan(avg):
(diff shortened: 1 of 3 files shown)


fallscameron01 (dismissed):
Nice work tracking these issues down. Ready to merge.

fallscameron01 (approved):
Nice catch.

PR #262, opened 2021-04-30
Create packages for all module algorithms

### LINKED ISSUE #246: Create __init__.py files for vision folders
## Description of Problem

Implement `__init__.py` files for vision folders to make importing vision algorithms easier.
See documentation for more information: https://docs.python.org/3/reference/import.html#regular-packages

## Requirements

- [x] Implement `__init__.py` files for all vision algorithm subdirectories used in competition.
- [x] Implement `__init__.py` file for vision directory.
- [x] Only import modules that are in use for competition in main vision directory.
- [x] Should be able to import entire folder with one import statement.
- [x] For example, `import module` would import all module algorithms.

mrouie:
- Packages were created for all competition-related algorithms using \_\_init\_\_.py files
- The function region_of_interest needed to be renamed to get_region_of_interest because the module name (filename) conflicted with the function name
Documentation and other naming were updated with the new function naming for get_region_of_interest
- The get_module_depth file needed to be renamed to module_depth for the same reason as region_of_interest
- Programs utilizing the aformentioned modules/functions were also updated with the new naming


fallscameron01 (approved):
Nice work. Looks ready for merge.

